PR #15, opened 2020-08-28
DOC: Add content to the lesson's overview page

jhlegarreta:
Add content to the lesson's overview page:
- Introduce the lesson and its target audience.
- Add the prerequisites.

### DIFF
--- a/index.md
+++ b/index.md
@@ -3,15 +3,31 @@ layout: lesson
 root: .  # Is the only page that doesn't follow the pattern /:path/index.html
 permalink: index.html  # Is the only page that doesn't follow the pattern /:path/index.html
 ---
-FIXME: home page introduction
+This Data Carpentry lesson aims to introduce learners to the analysis of
+diffusion Magnetic Resonance Imaging (dMRI) data using primarily Python. Its
+target audience includes researchers willing to discover the dMRI field, as well
+as under-graduate or graduate students willing to broaden their skills from
+related fields, such as neuroscience, computational neuroscience, or medical
+image analysis.
 
-<!-- this is an html comment -->
-
-{% comment %} This is a comment in Liquid {% endcomment %}
-
-> ## Prerequisites
+> ## Prerequisites for instructors
 >
-> FIXME
+> This lesson assumes that the instructors have solid knowledge about diffusion
+> MRI, scientific programming tools used throughout the lesson ([see the
+BIDS-dMRI Setup page](https://carpentries-incubator.github.io/SDC-BIDS-dMRI/setup.html)),
+and teaching experience.
+> If you are teaching this lesson in a workshop, please see the [Instructor notes](./guide/index.html).
+{: .prereq}
+
+> ## Prerequisites for learners
+>
+> This lesson assumes that the learners have a basic understanding of how a
+signal or image is represented digitally, an entry-level knowledge about aspects
+linked to medical image analysis, and a minimal set of computing and programming
+skills to install the required tools and to follow the lesson ([see the
+BIDS-dMRI Setuo page](https://carpentries-incubator.github.io/SDC-BIDS-dMRI/setup.html).
+This lesson requires learners to have completed the [Introduction to MRI and BIDS carpentry](https://carpentries-incubator.github.io/SDC-BIDS-IntroMRI/)
+corresponding to the Neuroimaging Curriculum.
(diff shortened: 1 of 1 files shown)


jhlegarreta:
@kaitj @josephmje do you know if there is a way to know if the page builds correctly/it renders as expected? Is there a CI that can be set up for the site?

The above PR assumes that a set up and instructor's guide or notes will be added in separate PRs in the near future. Shall we go on and add these? Maybe two initial PRs that can be later improved? Should it be specific to this lesson or shared across existing ones? Pinging @jerdra.

Also, I've observed that the `README` page's title, short explanation and  `About the lesson` paragraph are shared across the neuroimaging lessons. Is this intended?

josephmje:
There is no CI set up for the site but this could be an interesting way of previewing the pages. As of now, there are a couple different ways to build the site:
1.  If you have `ruby` and `bundler` installed on your computer ([instructions here](https://carpentries.github.io/lesson-example/setup.html)), you can run `make lesson-check` in the command line
2. If you don't have those software installed, you can also use a [Docker container](https://github.com/carpentries/lesson-docker) that includes them
3. You can enable GitHub Pages on your own fork so that the site builds at `<gh_username>.github.io/SDC-BIDS-dMRI`

jhlegarreta:
@josephmje thanks for the pointers. I am inclined to try the third option first and see what that gives.

jhlegarreta:
The third option does the job, and the PR contents are rendered correctly :+1:.

Maybe the CI option is something that is worthwhile considering/discussing at some point.

index.md
```diff
@@ -3,15 +3,32 @@ layout: lesson
 root: .  # Is the only page that doesn't follow the pattern /:path/index.html
 permalink: index.html  # Is the only page that doesn't follow the pattern /:path/index.html
 ---
-FIXME: home page introduction
+This Data Carpentry lesson aims to introduce learners to the analysis of
+diffusion Magnetic Resonance Imaging (dMRI) data using Python. Its target
```
kaitj, line 7: This looks good to me - only thing I would change is that the lesson primarily uses Python - there are some exceptions, mainly in the preprocessing (where FSL is used)

index.md
```diff
@@ -3,15 +3,32 @@ layout: lesson
...
+analysis.
 
-<!-- this is an html comment -->
-
-{% comment %} This is a comment in Liquid {% endcomment %}
+> ## Prerequisites for instructors
+>
+> This lesson assumes that the instructors have solid knowledge about diffusion
+> MRI, scientific programming tools used throughout the lesson ([see the
+BIDS-dMRI Setuo page](https://carpentries-incubator.github.io/SDC-BIDS-dMRI/setup.html)),
```
kaitj, line 17: Looks good - just spelling here: "BIDS-dMRI Setuo page"

kaitj:
> Also, I've observed that the `README` page's title, short explanation and `About the lesson` paragraph are shared across the neuroimaging lessons. Is this intended?

Not sure for the other neuroimaging lesssons. For the dMRI lesson, it had originally started off as a clone of one of the other two lessons, which is why it would share the same paragraphs. Was meaning to update after we had originally "finished" the lessons before we decided to update the lessons using ds000221.

index.md
jhlegarreta: Good catch. Fixed in 2f95b90.

index.md
jhlegarreta: Specified in 2f95b90. Thanks.

jhlegarreta:
> Not sure for the other neuroimaging lesssons. For the dMRI lesson, it had originally started off as a clone of one of the other two lessons, which is why it would share the same paragraphs. Was meaning to update after we had originally "finished" the lessons before we decided to update the lessons using ds000221.

OK, we'll figure this out once the lessons get to a more complete stage then !

josephmje:
> 
> 
> Overall looks great. Just a couple of minor things with the spelling and that python is primarily used with a few exceptions.
> 
> I guess one thing that I can't recall at the moment, are we making the "SDC-BIDS-IntroMRI" a prerequisite (@josephmje)? I know we have usually taught that first in the past.

Yes, I think we should make the IntroMRI course a prerequisite.

jhlegarreta:
> Yes, I think we should make the IntroMRI course a prerequisite.

Addressed in c134f4a.

PR #17, opened 2020-08-29
BUG: Rename eigenvalues and eigenvector images

jhlegarreta:
Rename eigenvalues and eigenvector images: the images and the filenames were
reversed.


jhlegarreta:
@josephmje @kaitj reviews are welcome.

kaitj:
Good catch! I might have updated the files incorrectly during the original reorganization of the repo. Looks good!

PR #23, opened 2020-09-14
ENH: Add `GitHub Actions` CI builds for notebooks

jhlegarreta:
Add `GitHub Actions` CI builds for notebooks.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -1,0 +1,43 @@
+name: Build, test
+
+on:
+  push
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout GitHub repository
+      uses: actions/checkout@v2
+
+    # Use a cache for the pip install to avoid reinstalling every time this is run.
+    - name: Cache pip install
+      uses: actions/cache@v2
+      id: cache
+      with:
+        path: .cache/pip
+        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+        restore-keys: |
+          ${{ runner.os }}-pip-
+    # Run with a Python version that we know works.
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.7'
+
+    # Install all requirements to run these tests.
+    - name: Install requirements
+      if: steps.cache.outputs.cache-hit != 'true'
+      run: |
+        python -m pip install --upgrade pip
+        pip3 install virtualenv
+        virtualenv ~/sdc_bids_dmri
+        pip install -r requirements.txt
+        pip install pytest nbval
+
+    # Run the actual tests of all the Python Notebooks.
+    - name: Test Notebooks
+      run: |
+        source ~/sdc_bids_dmri/bin/activate
+        pytest --nbval-lax -v *.ipynb
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # Introduction to MRI and BIDS
 
+![Build Status](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/actions?query=workflow%3A"Build%2C+test")
 [![Create a Slack Account with us][create_slack_svg]][slack_heroku_invite]
 [![Slack Status][slack_channel_status]][slack_channel_url]
 [![Binder][binder_svg]][binder_url]


jhlegarreta:
A few notes:
- I preferred to use `GitHub Actions` rather than using the `CircleCI` that I've seen is used by other lessons. IMHO, it allows to test a broader set of configuration (OS, etc.). Comment as appropriate.
- I have not added the status badge to the `README`. If we are satisfied with the `GitHub Actions` I will add it and push force.
- Other solutions to do this exist (e.g. [here](https://github.com/marketplace/actions/run-notebook#example-1---executing-notebook-with-parameters), [here](https://github.com/marketplace/actions/treebeard-jupyter-notebook-ci)). This is inspired by [this one](https://github.com/SimFin/simfin-tutorials/blob/master/.github/workflows/test.yml). I opted for not requiring a solution depending on the marketplace, but I honestly do not have strong arguments for/against any possibility, including using a docker image and skipping the `pip` installation, so suggestions are welcome.
- I locally run the `pytest` command and it indeed checks that the notebook can run. Eventually we will be willing to test the output cells, which means that we will need to save a checkpoint so that the run can be compared to a baseline I guess.
- Not sure how this will need to deal with the necessary input data. Unless all previous necessary data are stored, [chaining notebooks may be required](https://github.com/marketplace/actions/run-notebook#example-2---chaining-notebooks).
- This may require multiple PRs until we come to a successful solution.
- Do we want to test multiple Python versions? If lessons are launched using a docker image/binder may not be necessary, but still useful.
- Do we want to test multiple OS? Same rationale as before.
- The build does not perform PEP8 checks, but it could be added.
- Not sure why the CI build is not triggered for this very PR. Last time it happened to me on another repository I had to end up by adding the file directly through the GitHub interface and doing multiple PRs to see it working.

kaitj:
> * Not sure how this will need to deal with the necessary input data. Unless all previous necessary data are stored, [chaining notebooks may be required](https://github.com/marketplace/actions/run-notebook#example-2---chaining-notebooks).

For this, we do have the data available on OSF. Could we potentially choose 1-2 subjects to download and test? Taking a quick look at the GH action, it could just be the uncorrected input and the preprocessed output (with none of the command line tools installed). 

> * Do we want to test multiple Python versions? If lessons are launched using a docker image/binder may not be necessary, but still useful. 

Wouldn't be a bad idea to test multiple Python versions, especially if users are running this on their own. I think generally we'll stick with 3.x (with 2.x having reached end of life) though I am not sure which packages may only be supporting up to certain versions of 3.x. In terms of teaching, so far we have tried to set up a common environment, whether it be a binder instance or on a compute resource. 

> * Do we want to test multiple OS? Same rationale as before.

I think this one may require a little more thought. I am more familiar with the linux/unix environment nowadays. For any part of the lesson involving Python, this isn't a bad idea. However, I am not sure the command line tools run on Windows (maybe it does now with WSL). At the moment, I don't think we make any mention of OS requirements / limitations in the notebook, which may be an oversight? As mentioned in the previous comment, we've mostly been using a common environment currently. Something to think about.

> * The build does not perform PEP8 checks, but it could be added.

Probably a good idea - not even sure the notebooks as is would pass PEP8 checks.

> * Not sure why the CI build is not displayed here/triggered for this very PR. Last time it happened to me on another repository I had to end up by adding the file directly through the GitHub interface and doing multiple PRs to see it working. Maybe it is not displayed in here until merged into `master`, but the workflow is running in my fork: https://github.com/jhlegarreta/SDC-BIDS-dMRI/runs/1112110973?check_suite_focus=true

Yes, it would have to be merged into the master or there would have to be a branch in this repository with the changes. That said, should this be triggered on every push or just when a PR has been opened, reopened, or synchronized (changes pushed to that PR)? Could we assume the user would have tested locally and have an action trigger here on PR as a double check? I guess one benefit of having the action trigger this way is if the repository is forked, then any errors would be caught before a PR is made. Happy with it either way.

jhlegarreta:
Thanks for the comments.

> Could we potentially choose 1-2 subjects to download and test? 

If possible, I would start with a single subject, and see how it performs. Not sure how to do that now, though. Depending on the processing, completing the actions may be long or hit the time limit for our account I guess, so having tests on a minimal dataset of one subject would a good beginning.

> I think generally we'll stick with 3.x (with 2.x having reached end of life) though I am not sure which packages may only be supporting up to certain versions of 3.x

Personally, I would start with a 3.x that we know it works, have this merged if we manage to have the checks passing, then add different versions in separate PRs once we have the merged one as a baseline.

> For any part of the lesson involving Python, this isn't a bad idea. As mentioned in the previous comment, we've mostly been using a common environment currently.

Yes, something to think about at mid-term. For now, I would have the current/common `Linux` one working, then adding more versions and OS platforms can be done in separate PRs once a baseline is in place.

> Probably a good idea - not even sure the notebooks as is would pass PEP8 checks.

OK, will have this in mind. For now, I think we are better off having the checks without PEP8 checks, then adding them in a separate PR once we have hopefully checked and addressed issues locally.

> That said, should this be triggered on every push or just when a PR has been opened, reopened, or synchronized (changes pushed to that PR)? Could we assume the user would have tested locally and have an action trigger here on PR as a double check?

Safest would be having the action triggered for every push, which is the case now. As a developer, I prefer the checks to run on-line, rely on a common/known configuration, and free users from the burden of doing this locally.

kaitj:
> If possible, I would start with a single subject, and see how it performs. Not sure how to do that now, though. Depending on the processing, completing the actions may be long or hit the time limit for our account I guess, so having tests on a minimal dataset of one subject would a good beginning.

This sounds good to me. When you get a chance, maybe you can shoot me your email / username to add to the OSF repository. Long term, we may need to find a different storage solution. Need to take a second look at this with OSF changing to 50GB max per repo in November.

> Personally, I would start with a 3.x that we know it works, have this merged if we manage to have the checks passing, then add different versions in separate PRs once we have the merged one as a baseline.

This sounds good to me.

> Safest would be having the action triggered for every push, which is the case now. As a developer, I prefer the checks to run on-line, rely on a common/known configuration, and free users from the burden of doing this locally.

That is a great point.

I think all the concerns have been addressed. This PR is ready for merging unless there were other things that was mentioned in the discussion that we wanted to add?

kaitj:
@josephmje If this is okay with you, I will merge this in!

josephmje (approved):

Issue #31, opened 2020-11-04
BUG: `Github Actions` Test Failure


kaitj:
Good news is that the previously set up `Github Actions` now work. Bad news is it is failing when trying to install `pkg-resources==0.0.0` from the `requirements.txt`. 

Something to look into and fix to have valid tests!

josephmje:
we might need to clean up the `requirements.txt` file. it looks like the result of a `pip freeze` but we could pare it down to only the packages that are used in the lessons. also, some packages install others so those wouldn't need to be included.

jhlegarreta:
Thanks for the heads-up.

I did have the same issue when installing the packages locally and ended up by commenting that requirement.

On the one hand, I think having a fixed set of dependencies and versions ensures that the notebooks at least work with that set; on the other hand, if we get the CIs working and specify just the minimal set of packages that we explicitly import, that would be a guarantee that the notebooks work, and we would not need to rely on a `pip freeze`.

jhlegarreta:
PR #39 will hopefully address this or at least provide the appropriate fixes for the tests to be run.

jhlegarreta:
I think this can be closed: the builds are running the notebooks; the issues we face now (caching the data, the additional tools required, specifying the exact filename to the notebooks -i.e. solutions-, etc.) are being tracked in multiple threads.

PR #35, opened 2020-12-01
FIX: Trim Python package requirements

josephmje:
Addresses #31


jhlegarreta:
The change looks fair to me.

Not sure why the CIs are not being triggered, though. Not specific to this PR, since I realize that others have not triggered them; only merges to the main branch have triggered them, even thought they are set for `push` events and [should be enough](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#example-using-a-single-event). Maybe it has to do with triggering the GitHub Actions by pushes from forked repositories.

jhlegarreta (approved):

josephmje:
it looks like the install requirements step is now passing. but the test notebooks step is failing. does the path for the ipynb notebooks need to be updated?

jhlegarreta:
> it looks like the install requirements step is now passing. but the test notebooks step is failing. does the path for the ipynb notebooks need to be updated?

OK, I see the build in your fork. Probably the search pattern I put was relative the current directory and it should be adjusted. We could address that in a separate PR, and have this merged to make things go forward (and since in any case the downloading the data is a step that still requires to be added to the CI pipeline).

josephmje:
Sounds good! I'll merge this is and the CI pipeline can be updated separately.

PR #36, opened 2020-12-11
ENH: Add probabilistic tractography lesson

jhlegarreta:
Add probabilistic tractography lesson.


jhlegarreta:
Still need to check that the notebook runs.

josephmje:
is this ready for review @jhlegarreta ?

jhlegarreta:
I still need to make sure that this runs without issues, and will make the necessary fixes. Sorry for the delay. Will ping you here before the end of the year.

kaitj:
Would it make sense to change this PR to a draft until it is ready for review?

jhlegarreta:
> Would it make sense to change this PR to a draft until it is ready for review?

Done.

jhlegarreta:
There outputs show some artefacts in background regions. Not sure where the problem comes from. That would require some investigation, but for the time being I'd be for merging this.

kaitj (commented):
Looks good to me. One thought as I was reading through is there are a lot of acronyms, especially if someone reading this is new to diffusion / tractography. I wonder if it may be a good idea to just restate what the acronyms mean in acronym heavy paragraphs?

jhlegarreta:
> Looks good to me. One thought as I was reading through is there are a lot of acronyms, especially if someone reading this is new to diffusion / tractography. I wonder if it may be a good idea to just restate what the acronyms mean in acronym heavy paragraphs?

All are defined once, except for DTI and CSD, which are introduced in earlier notebooks/lessons. I would avoid restating them. Maybe having a sort of acronym/term cheat sheet shared across the notebooks could be envisaged at some point ?

josephmje:
> All are defined once, except for DTI and CSD, which are introduced in earlier notebooks/lessons. I would avoid restating them. Maybe having a sort of acronym/term cheat sheet shared across the notebooks could be envisaged at some point ?

I like the idea of having a glossary somewhere where we save all of the terms and definitions.

kaitj:
I should rephrase, it wasn't the number of acronyms, it was more the acronyms all seemed to pop up in the same 1-2 sentences at a glance. Anyway, I also like the idea of having a glossary somewhere.

jhlegarreta:
I opened an issue to keep track of the task of adding a glossary in a separate PR. So this is ready to be merged on my end.

_episodes/06-probabilistic_tractography.md
```diff
@@ -0,0 +1,363 @@
...
+
+where the curve $r(s)$ needs to be solved for.
+
+In probabilistic tracking, however, the tracking direction at each point along
+the path is chosen at random from a distribution of possible directions, and
+thus is no longer deterministic. The distribution at each point is different
+and depends on the observed diffusion data at that point. The distribution of
+tracking directions at each point can be represented as a probability mass
+function (PMF) if the possible tracking directions are restricted to a set of
+directions distributed points on a sphere.
```
josephmje, line 39:
```suggestion
distributed points on a sphere.
```

Is this what you meant here?

_episodes/06-probabilistic_tractography.md
```diff
@@ -0,0 +1,363 @@
...
+
+Like their deterministic counterparts, probabilistic tracking methods start
+propagating streamlines from a *seed map*, which contains a number of
+coordinates per voxel to initiate the procedure. The higher the number of seeds
+per voxel (i.e. the see density), the larger will be the number of potentially
```
josephmje, line 44:
```suggestion
per voxel (i.e. the see density), the larger the number of potentially
```

josephmje (approved):
Thank you for this contribution @jhlegarreta!

_episodes/06-probabilistic_tractography.md
jhlegarreta: Thanks for the review @josephmje :+1:. Changed in d77bd11.

_episodes/06-probabilistic_tractography.md
jhlegarreta: Addressed in d77bd11.

PR #39, opened 2020-12-17
FIX: Update ``.ipynb`` path and test on multiple Python versions

josephmje:
The `.ipynb` notebooks are being run now with the updated path but running into failures. Probably because we need to include a step to download the dataset and cache it.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -15,7 +15,7 @@ jobs:
     - name: Checkout GitHub repository
       uses: actions/checkout@v2
 
-    # Run with a Python versions 3.7, 3.8 and 3.9.
+    # Run with multiple Python versions.
     - name: Setup Python ${{ matrix.python-version }}
       uses: actions/setup-python@v2
       with:
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8]
 
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yaml
```diff
@@ -4,13 +4,23 @@ on:
...
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8]
 
     steps:
     - name: Checkout GitHub repository
       uses: actions/checkout@v2
 
+    # Run with a Python versions 3.7, 3.8 and 3.9.
```
jhlegarreta, line 18: I'd just say `Run with different Python versions`. That saves work if at some point a new version is added or another one is dropped.

jhlegarreta:
So rebasing this on `master` should now show the CI builds downloading the data.

.github/workflows/build-test.yaml
jhlegarreta: Done in 409c18b . Thanks Michael.

jhlegarreta:
So the notebooks are being run :rocket: !

We will need to install the additional missing dependencies (e.g. `fsl`), target the solutions for the notebooks for inline questions, create the output directories if not done within the notebooks, and see if other issues need to be fixed. But looking good overall !

In a separate PR we may want to cache the data so that it is downloaded only once for all tested Python versions.

josephmje:
Awesome! I'll merge this in. I haven't cached data before using GitHub Actions. @jhlegarreta would you like to take a crack at that?

jhlegarreta:
> Awesome! I'll merge this in. I haven't cached data before using GitHub Actions. @jhlegarreta would you like to take a crack at that?

Neither have I :grimacing:, but we'll figure that out :thinking:.

PR #41, opened 2020-12-20
ENH: Install additional dependencies for `GitHub Actions` tasks

jhlegarreta:
Install additional dependencies for `GitHub Actions` tasks.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -41,6 +41,14 @@ jobs:
         pip install -r requirements.txt
         pip install pytest nbval
 
+    # Install additional dependencies
+    - name: Install additional dependencies
+      run: |
+        wget https://fsl.fmrib.ox.ac.uk/fsldownloads/fslinstaller.py || echo "Failed to download FSL installer";
+        if [[ -f fslinstaller.py ]]; then
+          sudo python fslinstaller.py -d /usr/local/fsl --quiet || echo "Failed to install FSL";
+        fi
+
     # Download the necessary data
     - name: Download data
       run: |


jhlegarreta:
So looks like `FSL` gets installed. IDK if some environment variable will need to be configured to call its commands.

Other than that, some of the target notebooks need to target the solution files, and there are other issues in the log that need to be addressed. These can be done in separate PRs, and maybe that will allow us to know if the above is necessary (instead of adding an additional unnecessary step).

kaitj:
Looks good to me! Another package that will likely need to be included is ANTs for the registration in notebook 2 - can be included in another PR.

josephmje:
Ideally, I would prefer to avoid installing heavy dependencies like FSL and ANTS in our CI. I'm not sure what the best way to get around this would be. Possibly including the commands in Markdown rather than code cells and having some sort of visualization (screenshot or nilearn/dipy view of the file).

jhlegarreta:
> Ideally, I would prefer to avoid installing heavy dependencies like FSL and ANTS in our CI. I'm not sure what the best way to get around this would be. Possibly including the commands in Markdown rather than code cells and having some sort of visualization (screenshot or nilearn/dipy view of the file).

I thought they were required to test the pre-processing notebook. Thinking twice, if installing and building them takes a long time, we may hit the CI limit. But we will not know unless we try. If we do, we might need to build a docker container to pull in the CI.

josephmje:
> > Ideally, I would prefer to avoid installing heavy dependencies like FSL and ANTS in our CI. I'm not sure what the best way to get around this would be. Possibly including the commands in Markdown rather than code cells and having some sort of visualization (screenshot or nilearn/dipy view of the file).
> 
> I thought they were required to test the pre-processing notebook. Thinking twice, if installing and building them takes a long time, we may hit the CI limit. But we will not know unless we try. If we do, we might need to build a docker container to pull in the CI.

Yes, both FSL and ANTs are required for the pre-processing notebook. I'm all for trying it out to see how the CI does on time. What I was proposing was to maybe move those bash commands into a Markdown-formatted code block. We can save the intermediate results of those commands on OSF if they end up being used later.

Another solution might be to skip pytest for certain cells if that's an available option.

jhlegarreta:
> What I was proposing was to maybe move those bash commands into a Markdown-formatted code block. We can save the intermediate results of those commands on OSF if they end up being used later.

Looks fair. I do not see any inconvenient as of now to this; in fact, I believe these data is already available on OSF.

> Another solution might be to skip pytest for certain cells if that's an available option.

A quick search for those cells suggests adding `%%script false` or  `%%script false --no-raise-error` to skip the execution (untested). But that would also affect when teaching the lessons (should be removed manually if they are to run).

kaitj:
> Yes, both FSL and ANTs are required for the pre-processing notebook. I'm all for trying it out to see how the CI does on time. What I was proposing was to maybe move those bash commands into a Markdown-formatted code block. We can save the intermediate results of those commands on OSF if they end up being used later.

I wasn't even thinking about limits / time when commenting earlier so thanks for bringing this up. As @jhlegarreta mentioned, all of the intermediate data should be available in the OSF (I just can't remember if I uploaded single subject or a subset of 10 subjects).

I think I would be okay with moving these cells to a markdown cell (previous version of notebook 2 was all markdown). If this is the case, would it be helpful to have a link pointing to installation guide / downloads for FSL / ANTs in the notebook / in the pre-requisites?

jhlegarreta:
> I think I would be okay with moving these cells to a markdown cell (previous version of notebook 2 was all markdown). If this is the case, would it be helpful to have a link pointing to installation guide / downloads for FSL / ANTs in the notebook / in the pre-requisites?

Sounds fair.

kaitj (commented):
Missed this one too - thanks for adding this in.

Issue #29, opened 2020-11-03
Improve how the CSD reconstruction lesson screenshots are displayed


jhlegarreta:
Improve how the CSD reconstruction lesson screenshots are displayed:
- [X] Make the screenshots be displayed inline instead of launching a separate `FURY` window
- [x] Add additional screenshots corresponding to the anatomical orientations.

kaitj:
Just wanted to clarify what you meant by "corresponding to the anatomical orientations" so that I can add the screenshots you had in mind.

jhlegarreta:
> Just wanted to clarify what you meant by "corresponding to the anatomical orientations" so that I can add the screenshots you had in mind.

All current views are axial: adding coronal and sagittal views would provide useful visual information (leaving aside the fact that the tractograms can contain non meaningful streamlines).

kaitj:
Ahh that is fair. The current data is only a slice, so I'll turn it into a volume and show slices in each orientation. I also noticed it was using the original bvecs prior to eddy correction. Not sure if this was intended or if you had meant to use the eddy rotated bvecs. Either way, can be a quick change as i add in these screenshots.

jhlegarreta:
> Ahh that is fair. The current data is only a slice, so I'll turn it into a volume and show slices in each orientation

Have not checked the script lately; using only a slice was maybe done on purpose to avoid long waiting times. If this is the case, maybe choosing slices in the other two orientations would be enough?

> I also noticed it was using the original bvecs prior to eddy correction. Not sure if this was intended or if you had meant to use the eddy rotated bvecs. Either way, can be a quick change as i add in these screenshots.

Not intended.

kaitj:
Yep the slice was definitely intended to avoid long wait times, though a 30x30x30 volume seems to also run quick enough. And noted regarding the bvecs, Ive fixd and will push with the screenshot PR

PR #42, opened 2020-12-20
ENH: Make the `FURY` screenshots be displayed inline

jhlegarreta:
Make the `FURY` screenshots be displayed inline.

### DIFF
--- a/_episodes/05-constrained-spherical-deconvolution.md
+++ b/_episodes/05-constrained-spherical-deconvolution.md
@@ -88,6 +88,15 @@ the voxels with FA higher than a given threshold.
 from dipy.reconst.csdeconv import auto_response
 
 response, ratio = auto_response(gtab, data, roi_radius=10, fa_thr=0.7)
+
+# Create the directory to save the results
+out_dir = '../../data/ds000221/derivatives/dwi/reconstruction/sub-%s/ses-01/dwi/' % subj
+
+if not os.path.exists(out_dir):
+    os.makedirs(out_dir)
+
+# Save the FRF
+np.savetxt(os.path.join(out_dir, 'frf.txt'), np.hstack([response[0], response[1]]))
 ~~~
 {: .language-python}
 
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Partially addressed #29.

kaitj (changes requested):
Changes look good - just a few comments:

1. [Markdown] The basic equations summarizing SD: big gap between sentences, created presumably by the screenshot (doesn't necessarily need to be changed)
2. [Jupyter] The path for `t1_layout` includes an extra "33" just before `ds000221` in the first cell. (should probably be fixed).

jhlegarreta:
Thanks for the review Jason :+1:

> 1. [Markdown] The basic equations summarizing SD: big gap between sentences, created presumably by the screenshot (doesn't necessarily need to be changed)

Seems like Jekyll is unable to build the site and I cannot check how it looks like. The [file itself](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/fig/5/spherical_deconvolution_equation.png) does not seem to have an excessive white space around vertically. Maybe an issue can be opened and an screenshot added so that we can keep track of it separately?

> 2. [Jupyter] The path for t1_layout includes an extra "33" just before ds000221 in the first cell. (should probably be fixed).

Addressed in 1ffab3f.

josephmje:
> Seems like Jekyll is unable to build the site and I cannot check how it looks like. The [file itself](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/fig/5/spherical_deconvolution_equation.png) does not seem to have an excessive white space around vertically. Maybe an issue can be opened and an screenshot added so that we can keep track of it separately?

There was a mismatch with some of the file paths in the instructor guide. Fixed in #43 and was able to build the website fine locally.

josephmje (approved):

kaitj:
Haven't had a chance to look at the local website again, but am okay with merging given the pathing issue was addressed.

kaitj (approved):

PR #45, opened 2020-12-21
STY: Rename files and folders to increase consistency with markdown files

jhlegarreta:
Rename files and folders to increase consistency with markdown files.


jhlegarreta:
Assuming here that there is no `dash` vs `underscore` pattern convention that needs to be enforced by the Carpentries framework.

josephmje:
Thanks for this @jhlegarreta ! There is no naming convention for lessons in the Carpentries framework. That was just something that we came up with.

PR #46, opened 2020-12-24
ENH: Run GitHub Actions build-test workflow on PRs

jhlegarreta:
Run GitHub Actions build-test workflow on PRs.


jhlegarreta:
Build reports are now listed in the upstream repository.

kaitj:
Looks like the actions are running, just have to troubleshoot errors now. Quick glance, looks like top error related to the downloaded data location and where the notebook is looking for the data.

jhlegarreta:
Thanks for the heads-up @kaitj . Let's see if 22b5d3d makes the trick. Sharing the data across the jobs may come at a separate commit.

jhlegarreta:
Now data gets downloaded to `/home/runner/work/SDC-BIDS-dMRI/SDC-BIDS-dMRI/data`, which I'd dare to say it's where it should be; however, jobs still fail with the first error being
````
ValueError: BIDS root does not exist: /home/runner/work/SDC-BIDS-dMRI/SDC-BIDS-dMRI/data/ds000221
````
Not sure what's going on here.

jhlegarreta:
The data is now being fetched by the notebooks. That also involves more running time. Maybe at some point we will need to come up with ways to restrict the jobs to the notebooks that have been modified, if doing that is feasible, or limit the jobs to run e.g. once a week.

kaitj:
Another concern is do we know what the limits are for Github actions? From what I can find, job's are limited to 6 hours, and workflows 72 hours, but is there a limit per month? I agree that we probably need to come up ways to restrict the jobs a little bit.

jhlegarreta:
> Another concern is do we know what the limits are for Github actions? From what I can find, job's are limited to 6 hours, and workflows 72 hours, but is there a limit per month? I agree that we probably need to come up ways to restrict the jobs a little bit.

Yes, there is:
https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions

jhlegarreta:
> Both improvements involve more running time. And this is likely to increase as more notebooks are added. Maybe at some point we will need to come up with ways to restrict the jobs to the notebooks that have been modified, if that is feasible, or limit the jobs to run e.g. once a week.

If CIs are too long to run for every push/PR another possibility to consider is to run them e.g. only once a week using a scheduled (e.g. cron) job.

kaitj:
Sorry, think this one got lost in my list. I think, as you mentioned, less frequency of CIs are a good idea here especially given the run times.

kaitj (commented):
I'm not completely sure that this needs to be triggered on push. I like the idea, but we may be limited by resources (as has been discussed). I would think at a minimum, a trigger on PR is probably enough, again I think we've had brief discussions previously.

Anyways, not a major concern at this moment and the cron job was a good idea.

jhlegarreta:
> I'm not completely sure that this needs to be triggered on push. I like the idea, but we may be limited by resources (as has been discussed). I would think at a minimum, a trigger on PR is probably enough, again I think we've had brief discussions previously.
Anyways, not a major concern at this moment and the cron job was a good idea.

A trigger on PR can be misleading, since a first attempt may be successful/unsuccessful, and subsequent pushes in the same PR can change that result, and the status would be inconsistent with the actual behavior of the patch set.

So I think a weekly (or bi-weekly depending on the required time for the build) cron job would at least be a faithful and useful reflection of the repository's status. That might allow us enough build time to test the parts depending on the additional install dependencies.

I will try to adjust these things together with those in #41 at some point.

kaitj:
Just throwing this here so we don't forget: idea of using `[skip ci]` tags

jhlegarreta:
Add [Carpentries checks for lessons](https://carpentries.github.io/lesson-example/07-checking/index.html).

josephmje:
Can this be closed in favour of #126 getting merged?

jhlegarreta:
I think at term we'll be willing to test on a PR-basis. We still do not have a good/accurate estimation of how much time our tests need (when they succeed at least). And now I am unsure about the time limits for open source projects.

I'd keep it open to be rebased on `master`/`main` and allow builds on PRs once we are in a better position about testing our notebooks.

jhlegarreta:
Did not succeed in removing all but the two commits relevant to this PR. Superseded by PR #172.

PR #47, opened 2021-01-03
ENH: Re-organize lessons

jhlegarreta:
Re-organize lessons: re-group the local orientation reconstruction and
tractography lessons as different subsections of their topics, and provide an
overview/introduction section for them.

Update the lessons table accordingly in the `README`.

Fix the links to the lessons pages in the `README`.

Take advantage of the commit to update the `About the Lesson` section in the
`README` to match the actual contents.

PR #52, opened 2021-01-23
BUG: Remove duplicate probabilistic tractography episode

jhlegarreta:
Remove duplicate probabilistic tractography episode.

Take advantage of the commit to:
- Make the episode titles consistent.
- Fix the figure captions in the probabilistic tractography episode.


kaitj (commented):
Was wondering why there was a duplicate - thanks for deleting!

Being a little bit picky here, but wondering if it might be more accurate to say "Streamlines representing white matter" rather than "White matter streamlines"?

jhlegarreta:
> Was wondering why there was a duplicate - thanks for deleting!

The duplication was probably inadvertently introduced by me (sorry) when re-organizing lessons #47. Its contents were exactly the same as in [_episodes/09-probabilistic_tractography.md](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/_episodes/09-probabilistic_tractography.md).

Now, that led me to investigate a little bit more, since I remember a `04-tractography.*` file being in place a few months ago.
It is [still there as a notebook](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/code/04-tractography/04-tractography.ipynb), with its solutions, but it lacked the corresponding episode Markdown file.  I've gone through its contents, and I'd say that it fits into what I named as [08-deterministic_tractography.md](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/_episodes/08-deterministic_tractography.md). Looks an oversight on my end not to have moved its contents/renamed the file in #47.

So with some renaming to the notebook to make things consistent, some refactoring to (re)move part of the introduction in place/to [_episodes/06-tractography.md](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/_episodes/06-tractography.md), and some to [_episodes/07-local_tractography.md](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/_episodes/07-local_tractography.md), I think the lesson would already be in good shape !

A few more thoughts on that:
- Not sure why some (duplicate/deprecated) notebooks exist in the [files directory](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/tree/master/files). Shall we remove them in a separate PR?
- In order to avoid the need to rename all files every time these sort of things happen, or when a new episode is added (no the [figures folder](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/tree/master/fig) is also inconsistent with the current episode/notebook naming), we may want to adopt a convention as [suggested here](https://github.com/jhlegarreta/SDC-BIDS-dMRI/pull/1#issuecomment-762556306) in a future PR.

Also, in a separate PR we may need to stick to the 100-character line length format [recommended by the Carpentries](https://carpentries.github.io/lesson-example/04-formatting/index.html). I have used 80 consistently (at least in the markdown files), so maybe we are good using 80 all over the place. Looks like the Carpentries provide a script to check for this; interesting to be added to the CI in a future PR.

> Being a little bit picky here, but wondering if it might be more accurate to say "Streamlines representing white matter" rather than "White matter streamlines"?

Agreed. Done in 0cea529. Force pushed.

kaitj:
> Not sure why some (duplicate/deprecated) notebooks exist in the files directory. Shall we remove them in a separate PR?

We definitely should down the line once content is ready for delivery with the new dataset! They were kept in there during the transition as I think they were for the old dataset (`ds000030` I believe). I will also have to remember to remove that from the OSF so it is only `ds000221`.

> In order to avoid the need to rename all files every time these sort of things happen, or when a new episode is added (no the figures folder is also inconsistent with the current episode/notebook naming), we may want to adopt a convention as suggested here in a future PR.

Definitely open to it! I don't think there was any particular convention we were following previously.

> Also, in a separate PR we may need to stick to the 100-character line length format recommended by the Carpentries. I have used 80 consistently (at least in the markdown files), so maybe we are good using 80 all over the place. Looks like the Carpentries provide a script to check for this; interesting to be added to the CI in a future PR.

80 sounds good to me and yes I think it would be good to add to the CI! Unrelated, it may be good to add an `auto-assign` for reviewers if it passes the CI.

PR #54, opened 2021-01-24
DOC: Add terms to glossary

### LINKED ISSUE #44: Add a glossary of diffusion MRI terms
Add a glossary of diffusion MRI terms as a markdown cheat sheet. There exists a specific [reference.md](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/reference.md) page for this purpose.

Mentioned in PR #36.

jhlegarreta:
Add terms to glossary.

Adds the Carpentries [glossary component](https://carpentries.github.io/lesson-example/03-organization/index.html) as described [here](https://carpentries.github.io/lesson-example/reference.html) and resolves #44.

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -4,6 +4,77 @@ layout: reference
 
 ## Glossary
 
-FIXME
+~~~
+{:auto_ids}
+AD
+:   Axial Diffusivity
+
+BIDS
+:   Brain Imaging Data Structure
+
+CSA
+:   Constant Solid Angle
+
+CSD
+:   Constrained Spherical Deconvolution
+
+dMRI
+:   diffusion Magnetic Resonance Imaging
+
+DKI
+:   Diffusion Kurtosis Imaging
+
+DSI
+:   Diffusion Spectrum Imaging
+
+DTI
+:   Diffusion Tensor Imaging
+
+DWI
+:   Diffusion-Weighted Imaging
+
+EPI
+:   Echo Planar Imaging
+
+fODF
+:   fiber Orientation Distribution Function
+
+FA
+:   Fractional Anisotropy
+
+FOD
+:   Fiber Orientation Distribution
+
+FRF
+:   Fiber Response Function
+
+GFA
+:   Generalized Fractional Anisotropy
+
+MAPMRI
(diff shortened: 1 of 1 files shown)


kaitj (approved):

reference.md
```diff
@@ -4,6 +4,77 @@ layout: reference
...
+DKI
+:   Diffusion Kurtosis Imaging
+
+DSI
+:   Diffusion Spectrum Imaging
+
+DTI
+:   Diffusion Tensor Imaging
+
+DWI
```
josephmje, line 33: We probably only need either dMRI or DWI here.

reference.md
```diff
@@ -4,6 +4,77 @@ layout: reference
...
+GFA
+:   Generalized Fractional Anisotropy
+
+MAPMRI
+:   Mean Apparent Propagator MRI
+
+MD
+:   Mean Diffusivity
+
+ODF
```
josephmje, line 60: I don't know much about this but is there a big difference between the terms fODF, ODF and FOD?

reference.md
```diff
@@ -4,6 +4,77 @@ layout: reference
...
+SD
+:   Spherical Deconvolution
+
+SH
+:   Spherical Harmonics
+
+SHORE
+:   Simple Harmonic Oscillator Based Reconstruction and Estimation
+~~~
+{: .source}
```
josephmje, line 78: Will this work without `{: .source}`? Not sure if we want it to be displayed as code or regular text.

josephmje (approved):
Thanks for this @jhlegarreta! I just had a few minor comments/questions.

reference.md
jhlegarreta: I think we should keep both and explain in the introduction that they are used interchangeably, if not done. I personally prefer using dMRI (no specific reason).

reference.md
jhlegarreta: That would need to be explained in the local reconstruction episode I think.

reference.md
jhlegarreta: The referred Carpentries template uses it, if I got that right.

PR #65, opened 2021-03-14
MAINT: Update exercise solution

kaitj:
Update exercise solution to also use `ds000221` for dti episode

### DIFF
--- a/_episodes/04-diffusion_tensor_imaging.md
+++ b/_episodes/04-diffusion_tensor_imaging.md
@@ -212,12 +212,13 @@ Though other models are outside the scope of this lesson, we recommend looking i
 >> from nilearn import image as img
 >> import nibabel as nib
 >>
->> layout = BIDSLayout("../data/ds000030/derivatives", validate=False)
+>> deriv_layout = BIDSLayout("../data/ds000221/derivatives", validate=False)
+>> subj="010006"
 >>
->> t1 = layout.get(subject='10788', suffix='T1w', extension='nii.gz', return_type='file')[0]
->> dwi = layout.get(subject='10788', suffix='preproc', extension='nii.gz', return_type='file')[0]
->> bval = layout.get(subject='10788', suffix='preproc', extension='bval', return_type='file')[0]
->> bvec = layout.get(subject='10788', suffix='preproc', extension='bvec', return_type='file')[0]
+>> t1 = deriv_layout.get(subject=subj, space="dwi", extension='nii.gz', return_type='file')[0]
+>> dwi = "../data/ds000221/derivatives/uncorrected_topup_eddy/sub-%s/ses-01/dwi/dwi.nii.gz" % subj
+>> bval = "../data/ds000221/sub-%s/ses-01/dwi/sub-%s_ses-01_dwi.bval" % (subj, subj)
+>> bvec = "../data/ds000221/derivatives/uncorrected_topup_eddy/sub-%s/ses-01/dwi/dwi.eddy_rotated_bvecs" % subj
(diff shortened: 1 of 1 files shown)


_episodes/04-diffusion_tensor_imaging.md
```diff
@@ -212,12 +212,13 @@ Though other models are outside the scope of this lesson, we recommend looking i
...
->> layout = BIDSLayout("../data/ds000030/derivatives", validate=False)
+>> deriv_layout = BIDSLayout("../data/ds000221/derivatives", validate=False)
+>> subj="010006"
 >>
->> t1 = layout.get(subject='10788', suffix='T1w', extension='nii.gz', return_type='file')[0]
->> dwi = layout.get(subject='10788', suffix='preproc', extension='nii.gz', return_type='file')[0]
->> bval = layout.get(subject='10788', suffix='preproc', extension='bval', return_type='file')[0]
->> bvec = layout.get(subject='10788', suffix='preproc', extension='bvec', return_type='file')[0]
+>> t1 = deriv_layout.get(subject=subj, space="dwi", extension='nii.gz', return_type='file')[0]
+>> dwi = "../data/ds000221/derivatives/uncorrected_topup_eddy/sub-%s/ses-01/dwi/dwi.nii.gz" % subj
```
josephmje, line 219: Thanks for updating @kaitj! One thing I noticed is the string substitution. Do we use it anywhere else beforehand? If not, it might be a good idea to further explain it or use something else.

kaitj: I agree - I Believe it was also used (by me) in the preprocessing episode / notebook. I think at some point it would be good to go through and double check to make sure everything is consistent or explained.

jhlegarreta: I have used it in the CSD and prob tractography notebooks as well. IMHO the string substitution is just a way to save typing or to potentially change the subject at a single place, which I think is rather a programming resource/shortcut rather than something related to BIDS or the dMRI lesson, isn't it? I'd be for using it consistently. Otherwise, do you think there are better ways to do this?

jhlegarreta (approved):

_episodes/04-diffusion_tensor_imaging.md
josephmje: I agree, we should be consistent. Whatever method we choose, we should maybe just highlight it the first time it's used. My preference is f strings (eg. `f"../data/ds000221/derivatives/uncorrected_topup_eddy/sub-{subject}/ses-01 ...`). I think it's a little easier to follow than the `%s` substitution. But this is just being nitpicky.

jhlegarreta: I do not have a strong preference over f strings or substitutions. If we should go for f strings, an issue can be opened to keep track of that, or directly do a PR (separate from this one I'd say) to change them all at once. And maybe that would also mean finding and changing occurrences beyond the subject id substitution.

josephmje (approved):

_episodes/04-diffusion_tensor_imaging.md
kaitj: I think when this lesson was originally built out, I might have started it on python2.7, which may explain the substitutions. That said, my preference nowadays is also f strings, for the same reason as @josephmje.

jhlegarreta: Sounds reasonable ! Opened issue #66 to keep track of it and so that we can move forward and address it when we will find the time. Merging.

PR #59, opened 2021-02-08
ENH: Refactor tractography episodes

jhlegarreta:
Refactor tractography episodes:
- Add the deterministic tracking episode Markdown file.
- Make the deterministic tractography be consistent with the dataset and subject
  id used in the current version of the lesson.
- Use `matplotlib` to display the tractogram scene in the deterministic
  tractography episode and avoid opening the `fury` window.
- Add contents to the local tracking Markdown file.
- Improve the explanations in the probabilistic tracking episode.
- Refactor the tractography general Markdown file to make the contents more
  consistent across the related episodes.
- Remove the output error messages in the deterministic tracking Jupyter
  notebook.

Take advantage of the commit to:
- Remove the `{: .language-python}` episode Markdown-specific markup from the
  probabilistic tracking Jupyter notebook.
- Remove the episode Markdown-specific header from the probabilistic tracking
  Jupyter notebook.
- Add the subject id to the probabilistic tracking Markdown file and remove the
  unnecessary/unused call to `window.record`.
- Make the episode titles be consistent.
- Use the reference to the first work that proposed the global tracking method.
- Fix typos (e.g. `propogate` to `propagate`; `see density` to `seed density`).
- Use the term "episode" to refer to the sections.
- Use a double white space before inline comments to conform to PEP8.

PR #67, opened 2021-03-14
STYLE: Use the episode order feature

jhlegarreta:
Use the episode order feature.

Using this feature allows to override the ordering determined by the episode
filenames. This is also useful to avoid needing to rename all subsequent files
to keep their numbers consistent when a new episode has to be added somewhere in
the middle.

Following from the above:
- Remove the numbers from the filenames and folders.
- Rename the figure folders to match their corresponding episode names.

Take advantage of the commit to fix the figure relative location in some of the
Jupyter notebooks.


jhlegarreta:
Addresses what has been discussed several times (cf. [to this comment](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/59#issuecomment-775987883)).

kaitj (approved):
This looks good to me.

josephmje (approved):
Thanks @jhlegarreta! This change makes a lot of sense and will be easier for us to make changes in the future without having to constantly rearrange files and figures. 

Do you know if there's a way for the [figures page](https://deploy-preview-67--carpentries-dmri.netlify.app/figures/index.html) to respect the episode ordering?

jhlegarreta:
@josephmje Do you know if `netlify` reports somewhere any errors when building the site? The first version I pushed had a couple of them and was not building, and it looks like this made `netlify` unable to run. Thanks.

jhlegarreta:
> Do you know if there's a way for the figures page to respect the episode ordering?

Nice catch. I ignore if there is a way to do that. Maybe Toby will know.

josephmje:
> @josephmje Do you know if `netlify` reports somewhere any errors when building the site? The first version I pushed had a couple of them and was not building, and it looks like this made `netlify` unable to run. Thanks.

I just checked the netlify website and it looks like your previous version was still building. I think netlify is meant to report any build failures as comments on the PR. I'll look into seeing if we can impose a time limit to prevent a build from running for several hours.

jhlegarreta:
@josephmje Thanks. The failure has now been reported, but as you say, took a long time to complete.

jhlegarreta:
>> Do you know if there's a way for the figures page to respect the episode ordering?

> Nice catch. I ignore if there is a way to do that. Maybe Toby will know.

@tobyhodges Following what we [had discussed about episode ordering](https://github.com/jhlegarreta/SDC-BIDS-dMRI/pull/1#issuecomment-762556306), we've (renamed and) manually re-ordered episodes in this PR using the `episode_order` variable. However, we've realized that the `Figures` in the `Extras` section are ordered alphabetically matching the episode titles, and hence do not match the manual ordering.

I've been looking for a way to change this in order to match to the episode order set through the `episode_order` variable. As far as I've seen, the lesson example only provides a way to order the subsections in the `Extras` (i.e. `Reference`, `About`, `Discussion`, etc.). Do you know if there is a way to get the `Figures` ordered to match the episode ordering? Thanks.

tobyhodges:
Oh wow, that is a great point that I had not considered before! I will explore and hopefully get back to you with a fix soon...

tobyhodges:
Fixed in #68

PR #68, opened 2021-03-16
Manual episode order for Figures

tobyhodges:
This repository had an older version of `_extras/figures.md`, which pre-dates the introduction of manual episode ordering for lessons. This PR brings that file up-to-date with the most recent version of the file, and should cause your Figures page to respect the order of episodes in the lesson.

### DIFF
--- a/_extras/figures.md
+++ b/_extras/figures.md
@@ -3,11 +3,17 @@ title: Figures
 ---
 
 {% include base_path.html %}
+{% include manual_episode_order.html %}
 
 <script>
   window.onload = function() {
     var lesson_episodes = [
-    {% for episode in site.episodes %}
+    {% for lesson_episode in lesson_episodes %}
+      {% if site.episode_order %}
+        {% assign episode = site.episodes | where: "slug", lesson_episode | first %}
+      {% else %}
+        {% assign episode = lesson_episode %}
+      {% endif %}
     "{{ episode.url }}"{% unless forloop.last %},{% endunless %}
     {% endfor %}
     ];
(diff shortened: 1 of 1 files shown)


kaitj (approved):

josephmje (approved):
Thank you for this fix @tobyhodges!

jhlegarreta:
@tobyhodges as always, :100: ! Thank you.

Following your comment on the PR about the previous `figure.md` in this repository being out-of-sync, I've seen that the [examples `figure.md`](https://github.com/carpentries/lesson-example/blob/gh-pages/_extras/figures.md) is also different. Is there a place where we can look at the latest copies of the files? Is there a way to automatically have a lesson repository such as this one sync'ed with the upstream changes while keeping any overriding changes? Thanks.

josephmje:
> 
> 
> @tobyhodges as always, 💯 ! Thank you.
> 
> Following your comment on the PR about the previous `figure.md` in this repository being out-of-sync, I've seen that the [examples `figure.md`](https://github.com/carpentries/lesson-example/blob/gh-pages/_extras/figures.md) is also different. Is there a place where we can look at the latest copies of the files? Is there a way to automatically have a lesson repository such as this one sync'ed with the upstream changes while keeping any overriding changes? Thanks.

My guess is that they're coming from the [carpentries-theme](https://github.com/carpentries/carpentries-theme/blob/master/bin/boilerplate/_extras/figures.md) repository.

jhlegarreta:
> My guess is that they're coming from the carpentries-theme repository.

Looks the same file indeed. I also see that they match the [template lesson](https://github.com/carpentries-incubator/template) so this look good.

tobyhodges:
> My guess is that they're coming from the carpentries-theme repository.

That's absolutely right. The `_extras` files and other files that form the site content are not loaded in from that remote theme, but you can manually sync those up to get the latest version.

jhlegarreta:
> That's absolutely right. The _extras files and other files that form the site content are not loaded in from that remote theme, but you can manually sync those up to get the latest version.

@tobyhodges thanks for the explanation !

PR #71, opened 2021-03-19
STYLE: Prefer `{{ relative_root_path }}` to resolve figure paths

jhlegarreta:
Prefer the `{{ relative_root_path }}` placeholder to resolve figure paths.

Figures need to be located necessarily in a `fig` folder at the root level.
Using the placeholder allows to solve the paths without needing to use any
relative path and regardless of the location of the file containing it.

### DIFF
--- a/_episodes/local_tractography.md
+++ b/_episodes/local_tractography.md
@@ -23,7 +23,7 @@ Streamline propagation is, in essence, a numerical analysis integration
 problem. The problem lies in finding a curve that joins a set of discrete local
 directions. As such, it takes the form of a differential equation problem of
 the form:
-![Streamline propagation equation](../fig/local_tractography/streamline_propagation_diff_equation.png){:class="img-responsive"} \
+![Streamline propagation equation]({{ relative_root_path }}/fig/local_tractography/streamline_propagation_diff_equation.png){:class="img-responsive"} \
 Streamline propagation differential equation
 
 where the curve $r(s)$ needs to be solved for.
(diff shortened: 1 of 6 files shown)


tobyhodges:
> @tobyhodges Am I right in guessing that we prefer base to page.root for this:

Honestly this is one of the questions about the lesson template that I am still not sure about! It is definitely something that should be documented in the Lesson Example, but isn't. For now, if this is working for you, I would recommend to go with it and I will report back when I have a definitive answer for you.

tobyhodges:
But one note: I think the `base` variable was renamed to the more explicit `relative_root_path`. You may need to update that here.

jhlegarreta:
@tobyhodges sounds fair. Thanks for the explanation. Will update to `relative_root_path`.

josephmje (approved):
Looks like the figures still display correctly after this change!

EDIT: Except for the local tractography figure

_episodes/local_tractography.md
```diff
@@ -23,7 +23,7 @@ Streamline propagation is, in essence, a numerical analysis integration
 problem. The problem lies in finding a curve that joins a set of discrete local
 directions. As such, it takes the form of a differential equation problem of
 the form:
-![Streamline propagation equation](../fig/local_tractography/streamline_propagation_diff_equation.png){:class="img-responsive"} \
+![Streamline propagation equation]({{ relative_root_path }}/local_tractography/streamline_propagation_diff_equation.png){:class="img-responsive"} \
```
josephmje, line 26:
```suggestion
![Streamline propagation equation]({{ relative_root_path }}/fig/local_tractography/streamline_propagation_diff_equation.png){:class="img-responsive"} \
```

jhlegarreta: Thanks. Done in 765e93f.

jhlegarreta:
@josephmje I think the placeholder would also work in the Jupyter notebooks, but I'd leave that for a separate PR in case it does not and so as to prevent this improvement from being merged.

PR #73, opened 2021-03-19
STY: Sync template and update Makefile/Gemfile

josephmje:


jhlegarreta:
@josephmje The episode manual ordering introduced in #68 is being overriden/overwritten and is turned into alphabetical ordering :warning: .

jhlegarreta (changes requested):
@josephmje The episode manual ordering introduced in #68 is being overriden/overwritten and is turned into alphabetical ordering :warning: .

josephmje:
my mistake @jhlegarreta, thank you for catching this! i accidentally undid those edits when merging

jhlegarreta:
@josephmje Has netlify been able to build the latest commit?  There is no deployment message, right? Do you know if something is preventing the site from being built?

josephmje:
> @josephmje Has netlify been able to build the latest commit? There is no deployment message, right? Do you know if something is preventing the site from being built?

Yes, it's been built. netlify just updates the original deployment message

jhlegarreta (changes requested):
> Yes, it's been built. netlify just updates the original deployment message

:+1:

A few comments yet to be addressed, sorry:
- The `_config.yml` file is still being overridden (see the `workshop_repo:` variable and others)
- Our `.gitignore` file [has been modified](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/.gitignore) to disregard files generated by the editors we use, local site builds, etc. The current patch set is overriding these.

The file `bin/lesson_initialize.py` contains a reference to a file (`01-introduction.md`) that does not exist in our lesson (not sure if this should also be fixed in the upstream repository or how this affects lessons). Not sure about the purposes of the list, if all episodes should be listed should, etc. Have not investigated.

Thanks.

josephmje:
> A few comments yet to be addressed, sorry:
Done in [51d6146](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/commit/51d6146e27078fb37ddda3677e75552c059eb8dc).

> * The `_config.yml` file is still being overridden (see the `workshop_repo:` variable and others)
I kept the original variable for `workshop_repo` as I don't think this was meant to be changed. However, I noticed that this is being used in the setup instructions so I replaced it with `lesson-github`.

> * Our `.gitignore` file [has been modified](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/blob/master/.gitignore) to disregard files generated by the editors we use, local site builds, etc. The current patch set is overriding these.
My mistake, I've added them back in.


> The file `bin/lesson_initialize.py` contains a reference to a file (`01-introduction.md`) that does not exist in our lesson (not sure if this should also be fixed in the upstream repository or how this affects lessons). Not sure about the purposes of the list, if all episodes should be listed should, etc. Have not investigated.
I think this should be fine to keep. This script is only run with the lesson is first initialized . By default, an `01-introduction.md` file would've existed in the `_episodes` folder we've replaced that with `introduction.md`. It doesn't seem like this script is being run in the Makefile either which is good.

jhlegarreta (approved):
:+1: Thanks for doing this.

PR #76, opened 2021-03-20
ENH: Put the `fury` camera setup into a function

jhlegarreta:
Put the `fury`camera setup into a function.

Allows to reuse the code, avoids boilerplate code and improves readability.


jhlegarreta:
Should be possible to reuse it in the tractography notebooks.

kaitj (approved):
Thanks for adding this!

PR #77, opened 2021-03-20
DOC: Add discussion section about camera setup

jhlegarreta:
Add discussion section about camera setup.

### DIFF
--- a/_extras/discuss.md
+++ b/_extras/discuss.md
@@ -1,6 +1,30 @@
 ---
 title: Discussion
 ---
-FIXME
+
+## Setting up the camera viewpoint
+
+The [Constrained Spherical Deconvolution] episode defines an utility function
+(`generate_anatomical_views`) that receives a list of data and generates a
+[Matplotlib] `Figure` instance containing the axial superior, sagittal right
+and coronal anterior views of the data at the specified slices. The method
+achieves this by setting up the camera in the appropriate viewpoints. These
+viewpoints are obtained applying rotations at given angles; in the mentioned
+method, the rotations are applied to the focal point. These rotations are then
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Following the discussion in #69.

#76 should be merged before this.

This should probably be be cross-referenced from the episode and notebook. Not sure about the best/cleanest way to do it.

kaitj (commented):
This looks good to me. The sagittal view might actually be sagittal left. Was originally suppose to be sagittal right, and I think I never changed the comment. I can try to create a figure later this week to add to this, but agreed we should figure out if the image filepath is valid.

jhlegarreta:
> This looks good to me. The sagittal view might actually be sagittal left. Was originally suppose to be sagittal right, and I think I never changed the comment. I can try to create a figure later this week to add to this, but agreed we should figure out if the image filepath is valid.

I'd first merge #76, then do a separate PR to either change the title to sagittal left or generate a sagittal right view. But that is independent of the additions of this PR. I had similar conocerns about the coronal view being posterior (but did not had a detailed look at it): both left/right and anterior/posterior changes can be done in a separate PR if appropriate.

OK as for generating the figure as time permits.

kaitj:
> I had similar conocerns about the coronal view being posterior (but did not had a detailed look at it)

Yeah, I realized I never had a chance to double check orientations. I'll put it on my to do list!

kaitj:
Leaving this here more as a note for now so its not forgotten / can be double checked. 
Looks like two views that need to be changed are:
* Sagittal right -> Sagittal left
* Coronal anterior -> Coronal posterior

kaitj (approved):

PR #86, opened 2021-03-26
include base_path.html for relative_root_path

tobyhodges:
I noticed that your figures are not appearing on the lesson site. You are correctly using `relative_root_path` to specify the location of your image files, but this variable is only defined in the page if you `include` the `base_path.html` file in the page content. This PR adds that line to the beginning of every episode file, and should result in your images being displayed.

My next task is to document this in the lesson example...

### DIFF
--- a/_episodes/constrained_spherical_deconvolution.md
+++ b/_episodes/constrained_spherical_deconvolution.md
@@ -13,6 +13,8 @@ keypoints:
 - "It allows to resolve complex fiber configurations, such as crossings"
 ---
 
+{% include base_path.html %}
+
 ## Constrained Spherical Deconvolution (CSD)
 
 Spherical Deconvolution (SD) is a set of methods to reconstruct the local fiber
--- a/_episodes/deterministic_tractography.md
+++ b/_episodes/deterministic_tractography.md
@@ -12,6 +12,8 @@ keypoints:
 - "Deterministic tractography methods perform tracking in a predictable way"
 ---
 
+{% include base_path.html %}
+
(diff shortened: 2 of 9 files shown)


josephmje:
Thank you for this fix @tobyhodges! We were actually discussing this yesterday. The images were still getting displayed when looking at the netlify build or a local build but not on the actual lesson site.

josephmje (approved):

jhlegarreta:
@tobyhodges so nice that you got the time to take a look at our lesson :medal_sports: and notice that we had a problem ! As @josephmje says we also noticed but we had no knowledge of the missing `include`.

Also, we went ahead since netlify was correctly retrieving the images, and also checked later that the local build was retrieving them correctly.

As always, thanks Toby. Merging.

jhlegarreta (approved):

tobyhodges:
> @tobyhodges so nice that you got the time to take a look at our lesson

I do a bit of "lurking" on the GitHub notifications from all the lessons in the Incubator 😆 Good to see the images showing up now.

> Also, we went ahead since netlify was correctly retrieving the images, and also checked later that the local build was retrieving them correctly.

I wasn't aware that this could be an issue with the Netlify previews, but the inconsistency between local builds and what gets displayed on GitHub Pages is exactly why the `relative_root_path` variable was introduced. Unfortunately when you refer to a variable that doesn't exist, Jekyll substitutes nothing into the page and keeps on going with the page builds, which means that problems like this can be hard to diagnose.

PR #91, opened 2021-03-26
BUG: Use the right bvec file in prob tractography episode

jhlegarreta:
Use the right bvec file in prob tractography episode.


jhlegarreta:
Following https://github.com/carpentries-incubator/SDC-BIDS-dMRI/issues/29#issuecomment-799753626.

kaitj (approved):

PR #92, opened 2021-03-26
ENH: Reuse the anatomical orientation views generating method

jhlegarreta:
Reuse the anatomical orientation views generating method in the probabilistic
tractography episode.

Take advantage of the commit to split the output image cells from the next block
cells in the Jupyter notebook.

Set to `None` the default value of the colormap in the anatomical orientation
views generating method in order to avoid applying an undesired colormap to the
tractograms.


jhlegarreta:
The `ipynb` import is not working, and thus the notebook will not run successfully.

Have also tried to change the import (since in reality the csd notebook is in another folder), but my attempts have not been successful. You may want to have a look as time permits.

kaitj:
This looks good to me regarding the changes to reuse function.

josephmje (approved):

kaitj (approved):
Have not tested the code, but changes look good to me!

Do we want to update sagittal right to sagittal left as part of this PR or leave it as a separate PR?

jhlegarreta:
> Do we want to update sagittal right to sagittal left as part of this PR or leave it as a separate PR?

I'd leave it for a separate PR.

PR #90, opened 2021-03-26
BUG: Fix probabilistic tractography notebook

jhlegarreta:
Fix probabilistic tractography notebook: the Jupyter engine was unable to parse
the notebook since it was missing some newline/end of line characters (`,`), and
thus did not constitute a valid JSON entity.

PR #94, opened 2021-03-27
BUG: Fix deterministic tractography notebook

jhlegarreta:
Fix deterministic tractography notebook: the Jupyter engine was unable to parse
the notebook since it was missing an end of line characters (`,`), and thus did
not constitute a valid JSON entity.


jhlegarreta:
Similar to #90.

kaitj:
Fixes the regular notebook, but solutions notebook still down. Depending on what we plan on doing with the markdown / jupyter, thi may not be a problem.

jhlegarreta:
Thanks for the catch Jason. Fixed in 5393c4c.

Merging.

Issue #88, opened 2021-03-26
BUG: Github Actions Cache


kaitj:
First noticed in a different repository, but looking at the logs, it doesn't look like the cache is actually being used during the Github actions. Not a breaking issue, but probably something we want to look into.

kaitj:
With #153 and the successful runs with caching, can this issue be closed?

@jhlegarreta @josephmje

jhlegarreta:
Yes, the cache is working fine, so it can be closed. Thanks.

Issue #85, opened 2021-03-25
Automatically generate the Markdown episodes from the Jupyter Notebooks?


jhlegarreta:
The maintenance burden to keep all contents synchronized across episode Markdown files, Jupyter notebooks and the solution Jupyter notebooks is considerable. Whenever a change is made to an episode, inadvertently leaving behind one of the three parts is quite likely.

It may be worthwhile to explore how feasible is to use [Jupytext](https://jupytext.readthedocs.io/en/latest/index.html) to convert the notebooks directly into Markdown, and whether the layout/formatting looks as nice as when writing natively in Markdown.

kaitj:
Definitely an option to explore, especially if we want to continue to keep all 3. One thing to keep in mind is that solutions should be removed from the Jupyter notebooks, but retained in the solutions Jupyter notebook - having trouble envisioning a way to easily replace solutions to a more "fill-in-the-blank" method that was being used in fMRI and in previous versions of dMRI. I can see maybe completely removing the content within a coding cell, though not sure if that is an ideal solution.

> The Carpentries Markdown style recommend line breaks at 100 characters to keep a consistent style and improve readability. Not sure if that can be automated or enforced in case the Markdown files got generated automatically.

I think we can enforce this using something like [`markdownlint`](https://github.com/markdownlint/markdownlint)? Also if this needs to be enforced on the Python side, can use a python linter like [`black`](https://github.com/psf/black), though I am not sure if this works for notebooks.

jhlegarreta:
A note on the linters: the Python linter and Markdown linter will probably format the code differently, beyond some basic criteria, like the line length, so if we are successful in using them, we'll need to take into account that the Markdown vs. Python formatting will be different. But automatization on formatting is a highly desirable feature, so I think discrepancy is tolerable.

jhlegarreta:
> Also if this needs to be enforced on the Python side, can use a python linter like black, though I am not sure if this works for notebooks.

Looks like there are a number of tools for notebooks: [`jblack`](https://github.com/jupyter-black/jblack) , [`black-nb`](https://pypi.org/project/black-nb/), [`jupyter-black`](https://github.com/drillan/jupyter-black)

PR #96, opened 2021-03-27
ENH: Fix GitHub actions notebook test workflow

### LINKED ISSUE #31: BUG: `Github Actions` Test Failure

jhlegarreta:
Fix GitHub actions notebook test workflow

Fixes #31.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -28,9 +28,10 @@ jobs:
       id: cache
       with:
         path: .cache/pip
-        key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+        key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ hashFiles('requirements.txt') }}
         restore-keys: |
-          ${{ runner.os }}-pip-
+            ${{ runner.os }}-pip-${{ matrix.python-version }}-
+            ${{ runner.os }}-pip-
 
     # Install all requirements to run these tests.
     - name: Install requirements
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -44,7 +44,9 @@ jobs:
     # Download the necessary data
     - name: Download data
       run: |
-        osf -p cmq8a clone ./data
+        osf -p cmq8a clone ${{ github.workspace }}/data
+        mv ${{ github.workspace }}/data/osfstorage/* ${{ github.workspace }}/data
+        rm -rf ${{ github.workspace }}/data/osfstorage
 
     # Run the actual tests of all the Python Notebooks.
     - name: Test Notebooks
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -52,4 +52,8 @@ jobs:
     - name: Test Notebooks
       run: |
         source ~/sdc_bids_dmri/bin/activate
-        pytest --nbval-lax -v code/*/*.ipynb
+        pytest --nbval-lax -v code/introduction/solutions/introduction_solutions.ipynb
+        pytest --nbval-lax -v code/diffusion_tensor_imaging/solutions/diffusion_tensor_imaging_solutions.ipynb
+        pytest --nbval-lax -v code/constrained_spherical_deconvolution/constrained_spherical_deconvolution.ipynb
+        pytest --nbval-lax -v code/deterministic_tractography/deterministic_tractography.ipynb
+        pytest --nbval-lax -v code/probabilistic_tractography/probabilistic_tractography.ipynb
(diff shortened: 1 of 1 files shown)


jhlegarreta:
We will not know if this works until we merge it. And hopefully we can know by tomorrow if we merge it before.

After what we have discussed so far about the CI tests and cross referencing PRs #41 and #46, this PR proposed to:
- Leave the preprocessing notebook out for now due to our concerns about its dependencies on `FSL` and `ANTs`.
- Run the solution files for the notebooks having exercises.
- Run the tests once a week.

As far as I see, our notebooks do not have fill-in-the-gaps like exercises as it was the case in previous versions (cross referencing https://github.com/carpentries-incubator/SDC-BIDS-dMRI/issues/85#issuecomment-808214056). However, I've seen that exercises are only stated in the Markdown files, being the `code/*.ipynb` and `code/*/solutions/*.ipynb` files exactly the same. Not sure about the rationale behind (see the related comment in https://github.com/carpentries-incubator/SDC-BIDS-dMRI/issues/85#issuecomment-808214056), but that is something to be addressed in a separate PR.

Adding the non-solutions files to the CI tests is left for a separate PR.

If at some point we switch to fill-in-the-gaps like exercises:
- The corresponding cells will need to be manually marked as not to be executed in the non-solutions file provided that it does not interfere with the notebook execution when teaching (see https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/41#issuecomment-749064690).
- The solutions file will be run anyway.

If the weekly tests pass after this PR is merged:
- We will need to add the Carpentries checks (see https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/46#issuecomment-798954615) in a separate PR.
- Regardless of whether we are under the time limit bill:
  - We may want to investigate the cache mechanism across the Python versions used (#88)
  - We may want to add the preprocessing notebook to the tests. If this turns out into a too long test time, we may perform it only monthly.

Triggering tests on PRs might be considered once all the above has hopefully been addressed and we are still under the time limit bill.

So in summary, this PR is ready to be merged and the rest of the enhancements can come in separate PRs.

josephmje (commented):
Thanks for this @jhlegarreta. Could you try changing the cache key?
`key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}`

jhlegarreta:
@josephmje Done in 74da157.

josephmje (approved):

jhlegarreta:
Also, I've seen the CI output from yesterday:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/2262435829?check_suite_focus=true

We made progress (intro and DTI passing), but now the CSD fails due to a timeout. Cells have 30 minutes to do their job, which seem largely enough; especially for the one failing, which is the FRF display. That cell takes almost no time when running the notebook on the browser, but I've tested locally with the `pytest` command, and it shows the same behavior as in the CI. A few thoughts:

- I've done some debugging and the statements that trigger this behavior are the `window.snapshot` calls.
- Mark the relevant cells with `# NBVAL_SKIP`. Have tried locally and it works. Can be a distracting element: even if we aim for the solutions for now (including CSD and prob tracking eventually), at term we'll want to check the non-solution notebooks, so we'll end up by including it in the relevant cells. Also, it means that we should make statements related to the skipped cells (e.g. `scene.rm(response_actor)`) either dwell in the same cell or mark to be skipped as well, and splitting the cells that do some processing and visualization to be able to mark with `# NBVAL_SKIP` only those calling the visualization methods.
- I've tried locally with the `--nbval` flag instead of `--nbval-lax` (have put testing the CSD and followed by another notebook to a shell file and have called it): the second notebook gets tested regardless of the flag provided that there are no `XFAIL`s in the first noteook, i.e. **if  `# NBVAL_SKIP`s are added where appropriate.
- The execution does not follow because we are not allowing `pytest` to collect the notebooks by itself. We could try to ignore tests with the `--ignore`, e.g.:
```
pytest --nbval -v code/**/*.ipynb --ignore=code/preprocessing/**/*.ipynb
```
or
```
pytest --nbval-lax -v code/**/*.ipynb --ignore=code/preprocessing/**/*.ipynb
```
I've been unable to successfully exclude the preprocessing notebobok even.

We'd still need to mark the problematic cells with  `# NBVAL_SKIP` to see the tests passing.

Also, that would collect the tests in alphabetic order (i.e. CSD before any other one for the current episodes). Although not a problem, does not follow the lesson workflow.

So we're left with the following possibility to see the tests passing:
- Refactor the cells having some visualization part using `fury` to isolate such parts and skip them by adding the `# NBVAL_SKIP` flag to the visualization cells.

jhlegarreta:
> So we're left with the following possibility to see the tests passing:
> - Refactor the cells having some visualization part using fury to isolate such parts and skip them by adding the # NBVAL_SKIP flag to the visualization cells.

Cross-referencing #128.

PR #97, opened 2021-03-27
ENH: Add coronal and sagittal views to GFA plot

jhlegarreta:
Add coronal and sagittal views to GFA plot in probabilistic tractography
episode.

Take advantage of the commit to save the figure and display it in the episode.


jhlegarreta:
On a separate PR we may want to save and display the peaks as a qualitative measure to ensure that they are accurate.

josephmje (approved):

jhlegarreta:
If the GFA stopping threshold value gets changed the figure file will need to be changed.

PR #103, opened 2021-03-28
BUG: Update probabilistic tractography episode

kaitj:
This updates the tractography to use a `lmax` = 2 and a GFA threshold of .2 to enforce use of tensor-like model due to problems with tractography with `lmax` = 6. Figures also have been updated.


jhlegarreta (approved):
Thanks for doing this.

Issue #102, opened 2021-03-28
FIX: Update deterministic tractography episode


kaitj:
Content seems fine, but noticed it is using the unprocessed data. Will want to update this with any updates to figures.

PR #104, opened 2021-03-28
ENH: Update deterministic tractography episode

### LINKED ISSUE #101: Generate and add figures to the deterministic tracking episode
The deterministic tracking episode has currently no figures. Generating the relevant figures would allow learners to have some insight on what can be achieved through the episode, especially when reading the Markdown version.

kaitj:
Resolves #101 (partially) and #102 Updated the `BIDLayout` call to make use of the preprocessed data, which it wasn't previously doing. Also added in a couple of figures demonstrating the FA and the resulting tractogram. 

Took advantage of the commit / PR to remove the solutions notebook which currently wasn't being tested anyways.


jhlegarreta (approved):
Have only checked superficially, but looks good.

Thanks for doing this Jason.

It's interesting to see that the we always have an artifact outside the brain volume, as it is the case for the probabilistic tractography episode. Requires further investigation in separate PRs.

jhlegarreta:
The deterministic tractography notebook contains the output cell contents in the `master` branch: I believe they were introduced by commits in this PR.

kaitj:
Thanks for catching that - I keep forgetting to clear the outputs. Will update and make another PR.

PR #107, opened 2021-03-28
ENH: Define the SH order at a single location

jhlegarreta:
Define the SH order at a single location.

### DIFF
--- a/code/probabilistic_tractography/probabilistic_tractography.ipynb
+++ b/code/probabilistic_tractography/probabilistic_tractography.ipynb
@@ -128,7 +128,8 @@
    "outputs": [],
    "source": [
     "response, ratio = auto_response(gtab, dwi_data, roi_radius=10, fa_thr=0.7)\n",
-    "csd_model = ConstrainedSphericalDeconvModel(gtab, response, sh_order=6)\n",
+    "sh_order = 6\n",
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Looks like the last pass of in PR #103 did not apply the changes to the notebook. I'd merge this first, then make those changes in a separate PR.

kaitj:
Huh, thats odd. I wonder if I didn't save the notebook when I made the commit.

kaitj (approved):

code/probabilistic_tractography/probabilistic_tractography.ipynb
```diff
@@ -128,7 +128,8 @@
    "outputs": [],
    "source": [
     "response, ratio = auto_response(gtab, dwi_data, roi_radius=10, fa_thr=0.7)\n",
-    "csd_model = ConstrainedSphericalDeconvModel(gtab, response, sh_order=6)\n",
+    "sh_order = 6\n",
```
kaitj, line 131: This should be a 2, but otherwise looks good.

jhlegarreta: I prefer to leave that for a separate PR as mentioned in https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/107#issuecomment-808951637. I know it is a tiny change, but it has its own reason, so better to do it in a separate PR.

PR #108, opened 2021-03-29
FIX: Update Binder link

josephmje:


jhlegarreta:
Trying to launch binder, it is taking quite some time, like more than half an hour (I manually stopped after that). Not sure if something is missing.

jhlegarreta (approved):

PR #112, opened 2021-03-29
ENH: Bring 404 page from lesson example

jhlegarreta:
Bring 404 page from lesson example.

### DIFF
--- a/404.md
+++ b/404.md
@@ -1,0 +1,25 @@
+---
+layout: base
+root: .
+permalink: 404.html
+title: "Page not found"
+---
+
+# Oops! We can't find that page.
+{: style="text-align: center;"}
+
+> ## Our apologies!
+>
+> We can't seem to find the page you're looking for.
+> Try going back to the <a href="javascript:history.back()">previous page</a> or
+> navigate to any other page using the navigation bar above
+> {%- if site.kind == "lesson" -%} or the schedule below {%- endif -%}.
+> If you got here by clicking on a link in the
+> {%- if site.kind == "lesson" -%} lesson {%- else -%} workshop {%- endif -%},
+> please report this link to the
+> {%- if site.kind == "lesson" -%} lesson developers {%- else -%} workshop organizers {%- endif -%}.
+{: .caution}
+
+{% if site.kind == "lesson" %}
+  {% include syllabus.html %}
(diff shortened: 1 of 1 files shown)


kaitj:
Like the idea of the 404 page, but seems like it isn't loading the Carpentries template (at least in the netlify builds)

image

jhlegarreta:
Yes, looks like it's missing some parts of the `theme`. Will have a look as time permits.

404.md
```diff
@@ -0,0 +1,25 @@
+---
+layout: base
```
josephmje, line 2:
```suggestion
layout: page
```

404.md
```diff
@@ -0,0 +1,25 @@
...
+root: .
+permalink: 404.html
```
josephmje, line 4:
```suggestion
permalink: /404.html
```

josephmje (commented):
just left a few suggestions. not sure whether they will change the rendering or not.

jhlegarreta:
Thanks for the review @josephmje .

I tried building the site locally and the 404 HTML takes the theme correctly, so maybe it's some `netlify` artifact.
404_dmri_carpentries

kaitj (approved):
I am good with merging this in spite of Netlify not rendering. Did a cross-reference with the 404 page in the `SDC-BIDS-IntroMRI` repo and everything is the same. 

I suggest taking a quick look following merge to make sure the published page renders correctly

jhlegarreta:
We went ahead and merged this hypothesizing that it was maybe a `netlify` artifact what is shown in the [picture here](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/112#issuecomment-810214586).

It turns out that `netlify` was correct, e.g.
https://carpentries-incubator.github.io/SDC-BIDS-dMRI/diffusion_tensor_imaging2/index.html

I've only been able to find a single relevant commit to the `404` page in the examples repository:
https://github.com/carpentries/lesson-example/commit/d8f93f065ec7d4fb24c087b5d7ff3ddec63bedac

Looks like we are missing something elsewhere. @tobyhodges do you happen to have an intuition of what could be missing? Thanks.

jhlegarreta:
Cross-referencing #140.

PR #120, opened 2021-04-03
ENH: Update deterministic tractography

kaitj:
Related to #105.This adds an exercise to change stopping criteria to `BinaryStoppingCriterion` for tracking and a quick visualization exercise to change the color of streamlines. I think this is the simplest alternative for the stopping criteria suggested in #105. ACT may be good to add in the future as a challenging exercise or more advanced lesson.

~~Will mark ready for review once I figure out the rotations for the cameras so I can add a figure to the solution output.~~

PR #121, opened 2021-04-03
FIX: Fix slice orientations

### LINKED ISSUE #118: Fix the coronal and sagittal views of the peaks/tractograms
Visualizing the three orthogonal views of the peaks and the tractograms, it can be seen the sagittal right is indeed a sagittal left and that the coronal view is flipped upside-down flip (not sure if it is an anterior or posterior).

Note that we have used a sagittal right view in the matplotib scalar maps, so sticking to that view seems the most consistent choice.

Adjusting the roll, yaw and pitch parameters will probably make the trick.

kaitj:
Resolves #118 - flipping the rotations in the yaw & roll in the sagittal view seems to have resolved this. Views should now be "Sagittal right" and "Coronal anterior" The rotations applied here was used to produce the figure in #120.


jhlegarreta:
Checked locally. A few comments:
- Are we sure that the coronal anterior is a coronal anterior? Looking at it I'd say that it's a coronal posterior; looking at the sagittal, things get a little bit more confusing. 
- Fixing the coronal by setting the sagittal means that the fix would work if we call the methods in a given order. We should reset the camera in each method and then perform the appropriate changes to it so that it does not depend on the changes that were applied previously.

I think all concerned figures should be updated within this PR.

Also, I've realized about a mistake of mine in PR #92 and have submitted PR #122, so prior to generating all figures, this branch should be rebased on `master` once the latter PR gets merged. My bad. Sorry @kaitj .

kaitj:
>     * Are we sure that the coronal anterior is a coronal anterior? Looking at it I'd say that it's a coronal posterior; looking at the sagittal, things get a little bit more confusing.

Ahh yeah, took another look, you are right. 29e2e6b should fix this to coronal anterior (`pitch` changed to `-90`)

>     * Fixing the coronal by setting the sagittal means that the fix would work if we call the methods in a given order. We should reset the camera in each method and then perform the appropriate changes to it so that it does not depend on the changes that were applied previously.

Yeah, I noticed that that too. I think for now, for the purposes of generating the figures in the correct orientations, this is probably temporarily fine, but definitely want to look into resetting the camera as a more permanent solution.

> Also, I've realized about a mistake of mine in PR #92 and have submitted PR #122, so prior to generating all figures, this branch should be rebased on `master` once the latter PR gets merged. My bad. Sorry @kaitj .

No worries!

jhlegarreta:
I've merged #122. We can rebase this on `master` and hopefully get all figures generated correctly within this PR.

utils/visualization_utils.py
```diff
@@ -44,11 +44,15 @@ def generate_sagittal_right_view(scene, size=(600, 600), offscreen=True):
         A `window.Scene` instance seen from a sagittal right view.
     """
 
-    scene.yaw(-90)
-    scene.roll(90)
+    scene.yaw(90)
+    scene.roll(-90)
     scene.reset_camera()
```
jhlegarreta, line 49: Shouldn't the `reset_camera` be called prior to setting the yaw and roll properties?

utils/visualization_utils.py
```diff
@@ -44,11 +44,15 @@ def generate_sagittal_right_view(scene, size=(600, 600), offscreen=True):
...
     sagittal_scene = window.snapshot(scene, size=size, offscreen=offscreen)
 
+    # Reset scene
+    scene.roll(90)
+    scene.yaw(-90)
```
jhlegarreta, line 54: Should the `reset_camera` do this job?

kaitj: I thought it would given the description: `Reset the camera to an automatic position given by the engine.`, but it does not seem to. Rotating alone brings the scene out of the camera's FOV, and resetting the camera only seemed to recenter.

`reset_camera` was previously used, but new scenes was still affected by previous rotations.

jhlegarreta: :pensive:  OK. Thanks for the explanation Jason.

jhlegarreta (approved):

jhlegarreta:
Figures will need to be updated in a separate PR then.

kaitj:
Yes, I created a new issue to keep track of this. Merged this in as I had already rebased my local master and it would have gotten merged in with #130 otherwise.

PR #122, opened 2021-04-04
BUG: Fix the anatomical slice figure sorting

jhlegarreta:
Fix the anatomical slice figure sorting: the
`generate_anatomical_slice_views` method was returning the coronal anterior,
sagittal right and axial superior figures, but the
`generate_anatomical_slice_figure` method was mixing that order.

Take advantage of the commit to set the correct titles to the views.

This makes it consistent with the volume figures.


jhlegarreta:
Left behind in PR #92. My bad.

kaitj (approved):

PR #126, opened 2021-04-04
DOC: Add info about downloading single subject from OSF

kaitj:
Added a single subject zip file to the OSF. Using `clone` clones the entire repository, which may take a while, while using `fetch` only allows you to download a single file at a time. 

Added instructions to use `fetch` to download and unzip this single subject zip, which contains all the preprocessed data for `sub-010006`

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -169,7 +169,17 @@ $ osf -p cmq8a clone ./data
 {: .bash}
 
 Notebooks expect them to be placed in the `data` folder that exists in the root
-of the repository.
+of the repository. 
+
+Note the above command clones the entire repository, which may be quite large and
+take a while to download. Alternatively, data from a single subject is available
+and can be downloaded by running:
+~~~
+$ cd ./data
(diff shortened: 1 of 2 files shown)


jhlegarreta (commented):
The `GitHub` workflow should be changed accordingly to download the data corresponding to a single subject. That will also speed-up the time required to set up the environment prior to actually testing the notebooks.

kaitj:
Done in d01b2f8.

setup.md
```diff
@@ -169,7 +169,17 @@ $ osf -p cmq8a clone ./data
 {: .bash}
 
 Notebooks expect them to be placed in the `data` folder that exists in the root
-of the repository.
+of the repository. 
+
+Note the above command clones the entire repository, which may be quite large and
```
jhlegarreta, line 174: Should it be "Note that the (...)"?

jhlegarreta (approved):
Thanks for doing this Jason. Have not tested, but looks good. We'll not be able to see the CI test result until next Sunday; you may want to keep an eye of that.

PR #128, opened 2021-04-04
BUG: Fix call to visualization helper and skip fury visualizations in tests

jhlegarreta:
Fix call to visualization helper and skip fury visualizations in tests.


jhlegarreta:
Went ahead and did this.

Tested locally; CI builds should pass now, unless skipping tests is taken as a failure.

Also, the below shows the execution times without visualization. So definitely testing once a week, even when including the non-solution and solution notebooks, should be below the limits. More frequent tests might be affordable depending on whether the preprocessing one finally gets in.

csd_testing

det_tracking_testing

prob_tracking_testing

Part of fba4368 was an oversight in PR #92 (i.e. calling the appropriate function in the notebook).

@kaitj I'd like to merge this soon to avoid conflicts with your standing PRs. Thanks.

jhlegarreta:
As a follow-up, it dawned on me that `vtk 8.1.2` generates some known segfaults in its Python version, at least for Python 3.7. I am using Python 3.6 locally, but was using `vtk 8.1.2`: upgrading to `vtk 9.0.1` resolved the failure in cell 4 (and consequently the `XFAIL` in cell 5) locally. However,
- Removing all `# NBAL_SKIP`s to have the visualization cells will still fail because the `utils` package is still not found when trying to be imported in cell 9 (cross-referencing fba4368). I guess this could be overcome by finding a way to package the notebooks and the `utils` package. But,
- The CI installs the latest `vtk` available, which is `vtk 9.0.1`:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/2262435829?check_suite_focus=true#step:5:110
And thus there is something else going on with the CI testing.

In summary, for now the current PR as it stands is the best we can aim for.

kaitj:
> Removing all # NBAL_SKIPs to have the visualization cells will still fail because the utils package is still not found when trying to be imported in cell 9 (cross-referencing fba4368). I guess this could be overcome by finding a way to package the notebooks and the utils package. 

Regarding the `utils` package, this is potentially a problem while going through the lessons as well. I've worked around it by adding it to the `sys.path` when running the notebook, though it is not the recommended method. I can't recall at the moment, do virtual environments also search `PYTHONPATH` - can potentially add the utils folder there and update the `setup.md`. Easiest would probably be to package it up for setup. 

Otherwise, this PR looks good to me.

jhlegarreta:
> Regarding the utils package, this is potentially a problem while going through the lessons as well. I've worked around it by adding it to the sys.path when running the notebook, though it is not the recommended method. I can't recall at the moment, do virtual environments also search PYTHONPATH - can potentially add the utils folder there and update the setup.md. 

I have not modified my `PYTHONPATH` and have not run into issues when running the notebooks from my virtual environment, at least when being launched from PyCharm. I usually have the PyCharm options checked for adding the current working dir to the `PYTHONPATH` and hence I am not seeing the issue.

>  can potentially add the utils folder there and update the setup.md. Easiest would probably be to package it up for setup.

Either solution is fine for me: if you happen to find the way to package this nicely and fix the issue, please go ahead. If you think that requires investigation, we can leave it for a separate PR.

kaitj:
I think we can merge this for now and investigate the pathing separately, although I would prioritize trying to fix it!

kaitj (approved):

PR #130, opened 2021-04-05
BUG: Call to visualization module

### LINKED ISSUE #129: Fix utils.visualizations_utils.py path
Noted in #128, the module needs to be packaged or added to the `sys.path` for calls in both the actions workflow, as well as going through the notebooks locally.

kaitj:
This PR resolves #129 by adding the `utils` directory to the `PYTHONPATH`. Local testing seems to indicate this should fix the issue of calling the module.

image
(had a spelling mistake with the syntax error)

Updated the actions workflow accordingly, as well as the `setup.md`. Also took advantage of the PR to remove lingering files in the `files` directory from the old dataset and update the setup script in there to add utils to path and download single subject dataset.


jhlegarreta (changes requested):
Thanks for doing this.

This PR contains commits that do not belong to the PR; probably needs a clean rebase on master.

Not sure why `files/setup_tutorial` is updated instead of being removed. The relevant instructions now live in `setup.md`. I think none of the files within that folder are being used; the entire folder is probably no longer useful.

kaitj:
The choice of updating rather than removing was for running the necessary setup steps when teaching live. It was originally created when the lessons were taught in person. We could opt to remove it completely - thoughts @josephmje ?

As for the commits, yeah bit of a mess with having rebased local master - can look into rebasing with remote master later tonight

josephmje:
The `setup_tutorial` was a convenient utility for the learners to run all of the bash commands necessary for setting up the environment in one go. We used this before we had the `setup.md` file. I don't really have a preference on whether to keep it or not. I think it would be a "nice to have" but also adds duplication and another script that needs to be maintained should things change in the future.

jhlegarreta:
> I think it would be a "nice to have" but also adds duplication and another script that needs to be maintained should things change in the future.

I see. I think having duplications increases our chances to have out-of-sync contents; see how is being the process of having the markdowns, the notebooks and their solutions.

kaitj:
`setup_tutorial` is now removed in cdd582b and branch was rebased from `remote/master`

Agree about the maintenance aspect. I've removed the file, but kept the folder as it was part of the initial carpentries template.

jhlegarreta (approved):
:+1:

Issue #131, opened 2021-04-05
Update figures


kaitj:
With slice orientations fixed (#121), figures will need to be updated accordingly. Episodes that make use of the utils call include
- [x] CSD
- [x] Deterministic tractography
- [x] Probabilistic tractography

kaitj:
Am I missing any other notebooks that use the call @jhlegarreta ? Those were the ones that I could find.

jhlegarreta:
> Am I missing any other notebooks that use the call @jhlegarreta ? Those were the ones that I could find.

Those are the ones.

PR #134, opened 2021-04-07
ENH: Update probabilistic tractography notebook 

kaitj:
PR updates the figures in the probabilistic notebook using the `utils` module (see #131) and also fixes the call to `generate_anatomical_volume_figure` in the notebooks. Outputs have been left in the notebooks (can be removed) given there are no exercises here and in previous notebooks, only the solutions were removed. 

In addition, static images and some markdown calls within the notebooks have been removed. Also took advantage of having run the notebooks to also lint code cells.


jhlegarreta:
Not sure about the coronal anterior being really anterior yet having a look at the MCP, but anyways.

As for the output, I'm fine leaving them for now; we'll want to keep them only in the eventual solutions notebook.

I see that we have some warnings due to some API changes; we may want to open the corresponding issues to keep track of them, and may need to check that they are present when using the versions the CIs/binder are using.

Feel free to merge.

kaitj:
Merging - will keep an eye on the coronal anterior image, may have accidentally rotated the wrong way when I went to reset the scene.

PR #139, opened 2021-04-17
FIX: add skip tag for utils

kaitj:
Adds the `#nbval_skip` tag to the cell causing failure for the workflow in the deterministic notebook. Also take advantage of commit to remove a local path from one of the cells.


jhlegarreta (dismissed):

jhlegarreta (changes requested):
@kaitj I'm thinking that instead of skipping the whole solution cell, it might be better to split the solution into cells that do the necessary computations from the one that does the visualization part. Otherwise the solution does not get tested. Thanks.

kaitj:
Split the visualizations in the solutions notebook in de36551. Also used the commit to add the skip tag for exercise 2.

jhlegarreta (approved):
Warnings will need to be addressed at some point.

PR #140, opened 2021-04-20
Update styles 2021-04-20

zkamvar:
This brings in updates from the styles repository, including fixes to the 404 page (https://github.com/carpentries/styles/pull/578)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,7 @@ trim_trailing_whitespace = true
 indent_size = 2
 indent_style = space
 max_line_length = 100  # Please keep this in sync with bin/lesson_check.py!
+trim_trailing_whitespace = false  # keep trailing spaces in markdown - 2+ spaces are translated to a hard break (<br/>)
 
 [*.r]
 max_line_length = 80
--- a/.github/workflows/template.yml
+++ b/.github/workflows/template.yml
@@ -1,0 +1,138 @@
+name: Test template
+on:
+  push:
+    branches: gh-pages
+  pull_request:
+jobs:
+  check-template:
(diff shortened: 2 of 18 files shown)


josephmje (changes requested):
Thanks for this @zkamvar! Looks like there's been a lot of updates to the styles repository.
Looks good to me. The only bit that I think shouldn't apply to this lesson are the extra GitHub Actions workflows. Could you please get rid of those?

jhlegarreta (approved):
@zkamvar thanks for doing this :100:.

@josephmje I think the GitHub actions are OK; although opposite to `netlify` they do not allow us to have a preview of the website, I was asking @tobyhodges about these and other checks in a private mail. So I'd keep them. Is there any reason you would not have them?

Also, I see no conflict or override to any local customization, so to me this is ready to be merged.

josephmje:
> @josephmje I think the GitHub actions are OK; although opposite to `netlify` they do not allow us to have a preview of the website, I was asking @tobyhodges about these and other checks in a private mail. So I'd keep them. Is there any reason you would not have them?

I was concerned about the website rendering bit and pushing to the `gh-pages` branch but it looks like that gets skipped since we don't have that branch. Fine to merge with me.

jhlegarreta:
> I was concerned about the website rendering bit and pushing to the gh-pages branch

Ah, sounds fair :+1:. I think that at some point we may need to rethink about it (e.g. automatically pushing to the `gh-pages` branch using a scheduled action, etc.). Probably needs to be discussed with the carpentries folks to also explain why we decided to use the `master` branch 🤔 .

>  but it looks like that gets skipped since we don't have that branch. Fine to merge with me.

Merging.

josephmje (approved):

PR #144, opened 2021-04-21
ENH: Add website build workflow status badge to `README`

jhlegarreta:
Add website build workflow status badge to `README`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # Introduction to dMRI
 
 [![Build Status](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/workflows/Build,%20test/badge.svg)](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/actions?query=workflow%3A"Build%2C+test")
+[![Website Check Status](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/workflows/Website/badge.svg)](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/actions/workflows/website.yml?query=workflow%3AWebsite)
 [![Create a Slack Account with us][create_slack_svg]][slack_heroku_invite]
 [![Slack Status][slack_channel_status]][slack_channel_url]
 [![Binder][binder_svg]][binder_url]


jhlegarreta:
Not sure why no status is being reported;  have tried creating the badge with different branches in the workflow itself to no avail. The badge is being displayed correctly in the `lesson-example`. Might have to do with https://github.com/carpentries/styles/issues/581.

josephmje:
> Not sure why no status is being reported; have tried creating the badge with different branches in the workflow itself to no avail. The badge is being displayed correctly in the `lesson-example`. Might have to do with [carpentries/styles#581](https://github.com/carpentries/styles/issues/581).

I think we want to display the status of the build-test workflow instead of the website workflow?

jhlegarreta:
> I think we want to display the status of the build-test workflow instead of the website workflow?

The build-test is already being displayed. The `website` check reflects the website build status (i.e. whether the markdown files can be process to be rendered, whether they stick to the carpentries style, etc.).

josephmje:
> > I think we want to display the status of the build-test workflow instead of the website workflow?
> 
> The build-test is already being displayed. The `website` check reflects the website build status (i.e. whether the markdown files can be process to be rendered, whether they stick to the carpentries style, etc.).

My mistake!

kaitj (commented):
This looks fine to me with regards to adding the badge.

Regarding the `No status`, could it be similar to this: https://github.community/t/workflow-badge-always-no-status-when-limited-to-branch-master/17087? 

Also from what I can tell, the `Build, test` status badge has only tested off this repo in recent runs - can't recall if the badge displayed correctly before. Similarly, last few workflow runs in `lesson-example` has also been on the main repo. Might be something along the lines of trying to query an actions workflow triggered by a branch on a fork? 

I'd be okay with merging this in, seeing what happens and opening a separate PR if changes are needed.

kaitj (approved):

jhlegarreta:
> Regarding the `No status`, could it be similar to this: https://github.community/t/workflow-badge-always-no-status-when-limited-to-branch-master/17087?

We are not limiting the badge to the master branch, which OTOH I think we should (my bad) and should in any case not be an issue.

> Also from what I can tell, the` Build, test` status badge has only tested off this repo in recent runs - can't recall if the badge displayed correctly before.

It has always displayed the expected status.

> Similarly, last few workflow runs in `lesson-example` has also been on the main repo. Might be something along the lines of trying to query an actions workflow triggered by a branch on a fork?

I'd say that it does not have to do with that. Inspecting more closely the workflow file, turns out that only `gh-pages` and `main` branches get tested when pushing, which we do not have. But `gh-pages` is created within the workflow and thus the workflow is run, but then no status is reported I guess. So maybe we should at least switch our `master` to `main`.

PR #153, opened 2021-05-06
FIX: Update Github Actions caching

josephmje:
I updated the "Build, test" workflow by removing the virtualenv creation step. I also set it to run only if changes are made to the `.ipynb` files

Checkout the last successful run [here](https://github.com/carpentries-incubator/SDC-BIDS-dMRI/actions/runs/817037226).

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -27,22 +27,18 @@ jobs:
       uses: actions/cache@v2
       id: cache
       with:
-        path: .cache/pip
-        key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ hashFiles('requirements.txt') }}
+        path: ~/.cache/pip
+        key: python-${{ matrix.python-version }}-${{ hashFiles('requirements.txt') }}
         restore-keys: |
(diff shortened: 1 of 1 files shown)


.github/workflows/build-test.yaml
```diff
@@ -27,19 +26,16 @@ jobs:
       uses: actions/cache@v2
       id: cache
       with:
-        path: .cache/pip
-        key: ${{ runner.os }}-pip-${{ matrix.python-version }}-${{ hashFiles('requirements.txt') }}
+        path: ${{ env.pythonLocation }}
```
josephmje, line 29:
```suggestion
        path: .cache/pip
```

.github/workflows/build-test.yaml
```diff
@@ -1,8 +1,7 @@
 name: Build, test
 
 on:
-  schedule:
-    - cron: "0 2 * * 1,3,5"  # 02:00 UTC every Mon,Wed,Fri
+  push
```
josephmje, line 4:
```suggestion
  push:
    paths:
    - '**.ipynb'
```

PR #154, opened 2021-05-07
STYLE: Bump style changes

jhlegarreta:
Bump style changes.


josephmje:
Thanks @jhlegarreta! Before merging, I just want to confirm whether the `_includes`, `_layouts` and `assets` folders should be included in the sync (see [here](https://github.com/carpentries/styles/issues/598)).

jhlegarreta:
>  Thanks @jhlegarreta! Before merging, I just want to confirm whether the _includes, _layouts and assets folders should be included in the sync (see here).

> Thanks for caring about those.

@josephmje please comment about the above folders when you have the information. If those files are to be excluded, maybe the relevant Carpentries repositories/materials should explicitly tell about it.

josephmje:
I got some clarification from @zkamvar in the Carpentries slack. I'll post his comments below:

Michael:
Would anyone happen to know the difference between the carpentries' styles repo and the carpentries-theme repo? To me, it seems like most of the development is happening in styles, which later ends up getting synced with carpentries-theme. However, the theme repo also seems to have some unique files that add to the way incubator lessons get displayed

Zhian:
I have a bit more context for this:
The mix of remote and non-remote themes are largely due to a combination of a rapid expansion of lessons on top of a model that was meant for a handful (the styles repo), the historical lack of support for remote themes on GitHub (until well into the expansion phase), and the delay between scale and additional staff to support development and maintenance (that's me!).
It's worth noting that the lesson infrastructure has gone through a lot of shifts historically (raw HTML -> pandoc -> Jekyll), and in the early days, those shifts have sometimes been rapid (https://software-carpentry.org/blog/2015/07/pushing-back.html).
The shift to remote themes occurred because it was seen as a way to modernize the infrastructure and make it "easier" to maintain the changes since lessons that use the remote theme would rely on a single source of truth for the styling, but it made a manual dependency chain longer and more difficult because the tools for building the lessons were embedded within the repository.
To be clear, we are absolutely moving away from this model as we described in the design principles for the next iteration of the lesson template (which is currently in alpha testing and should have a release candidate by the end of the summer): https://carpentries.org/blog/2020/08/lesson-template-design/

The maintenance chain right now is styles -> carpentries-theme -> incubator-theme

Michael:
Thanks for the thorough explanation Zhian! I really appreciate it. That makes a lot of sense to me. From reading the blog post, it looks like the future goal is to remove styling elements from the lesson repository. So me trying to automate syncing with the styles repo is probably not necessary at this point

Zhian:
Correct. Plus, I'll be working on the automation process this week. There were a few false starts earlier due to the shifting sands of GitHub workflow permissions (which I'm glad they have finally been locked down).

josephmje:
For this lesson, since we are using the `carpentries-theme` repo as a remote, it's fine to remove the `_includes`, `_layouts` and `assets` folders since those will be checked upstream. We should also remove any existing files we have in those folders and maybe only keep `_includes/links.md` and `_includes/lesson_links.md` since those files have been modified from the theme repo.

jhlegarreta:
@josephmje done in 5c8cb2d .

josephmje (approved):

PR #157, opened 2021-05-20
STYLE: Use `inline` image class

jhlegarreta:
Use `inline` image class to display images within a paragraph.


jhlegarreta:
Made possible thanks to https://github.com/carpentries/styles/pull/602.

maxim-belkin:
Yes, the remote theme repository has to be sync'd with carpentries/styles before you can use inline images.

josephmje (approved):

Issue #14, opened 2020-08-21
DOC: Convert Episode 1 to Jupyter notebook


kaitj:
A couple of things to update for episode 1:
  
- [ ] Need to update the subject in the episode from 010002 to 010006 (which is actually in the subset we are using)
- [ ] Also need to update exercises to use a different subject in the subset
- [x] The jupyter notebook (located in `code`) should be updated to reflect the changes of the new dataset. These changes were only made to the episode

PR #165, opened 2021-07-09
STYLE: Improve style

jhlegarreta:
Improve style:
- Change NBSP appearances to regular whitespaces
- Add missing closing bracket

PR #166, opened 2021-07-09
BUG: Fix introduction episode non-solution notebook

jhlegarreta:
Fix introduction episode non-solution notebook: use the correct dataset
and subject for the analyses.

Fixes:
```
ValueError: BIDS root does not exist: /home/runner/work/SDC-BIDS-dMRI/SDC-BIDS-dMRI/data/ds000030
```

raised in
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/3025093598?check_suite_focus=true#step:7:44

Synchronize the markdown episode contents and the Jupyter notebooks
(both the non-solution and solution).

Take advantage of the commit to:
- Fix the static image relative path.
- Fix the dataset tree: change the duplicate `fmap` folder for the
missing `func` folder.
- Add the `bids` initial dot extension config statement to avoid the
related warning being displayed.
- Remove the `dipy` and `fury` installation steps: both tools are
assumed to have been installed during setup.
- Use the same DWI volume for to be displayed.
- Fix the output of the bvecs.
- Removed the unexplained `b0` volume slicing block.
- Add exercise to notebooks.
- Fix the suffix syntax in exercise.
- Remove the second exercise: it is already explained/done at the
beginning of the episode.
- Fix typos.
- Remove outputs from non-solution notebook.


jhlegarreta:
Partially fixes #14 (addresses the third task).

I ignore if the `osf` copy of the data is a faithful mirror of the `openneuro` data; i.e. if the treee would look as it is written. I sticked to what I saw in `openneuro`. Please check and amend the commit as necessary.

I'd say that I'd be better to merge this prior to #165, since the latter's changes can more easily be applied if the rebasing derives in hard-to-solve conflicts.

kaitj:
Looks fine to me, I don't recall if the subject is still included in `ds000221`, but the structure would still be the same (just different subject id). This seems like a good intermediate step - if i recall @josephmje was working on a revamped intro (or was that already merged in)?

_episodes/introduction.md
```diff
@@ -122,18 +129,23 @@ tree '../data/ds000221'
...
-Lets first pull the metadata from its associated JSON file using the
+Let's first pull the metadata from its associated JSON file using the
 <code>get_metadata()</code> function for the first run.
 
 ~~~
+import bids
 from bids.layout import BIDSLayout
 
+?BIDSLayout
+
```
josephmje, line 140:
```suggestion
```

josephmje (approved):

_episodes/introduction.md
jhlegarreta: I left it because it shows the in-notebook help of the class, and thought that it was useful.

jhlegarreta:
Further improvements can come in separate PRs. This would be a step towards making the episode and its notebooks consistent, and hopefully have the tests passing, since the non-solution notebook is non-functional as it is now.

PR #169, opened 2021-07-23
ENH: Allow initial dot for BIDS extension entity

jhlegarreta:
Allow initial dot for BIDS extension entity.

Fixes:
```
FutureWarning: The 'extension' entity currently excludes the leading dot ('.').
As of version 0.14.0, it will include the leading dot.
To suppress this warning and include the leading dot,
 use `bids.config.set_option('extension_initial_dot', True)`.
  warnings.warn("The 'extension' entity currently excludes the leading dot ('.'). "
```


jhlegarreta:
> @jhlegarreta Is this change necessary? I remember originally, pybids required you to put in the initial dot when using the extensions argument but they've since migrated away from that. Their tutorials also don't use the initial dot. Could we instead query for "nii.gz" and skip the warning?

@josephmje file extensions were being queried without the initial dot across all notebooks before this PR, and were kept as they were; the warning was present, though. Please try it on your end with. I'm using `pybids 0.12.4` locally, same version as in our requirements. Thanks.

josephmje:
> @josephmje file extensions were being queried without the initial dot across all notebooks before this PR, and were kept as they were; the warning was present, though. Please try it on your end with. I'm using `pybids 0.12.4` locally, same version as in our requirements. Thanks.

My mistake, the change is actually going to other way and asking to include the initial dot. Should we make this change instead rather than dismissing the warning?

josephmje (approved):

PR #172, opened 2021-07-31
ENH: Run GitHub actions build test workflow on PRs

jhlegarreta:
- ENH: Run GitHub Actions build-test workflow on PRs
- ENH: Update Python versions in CI build matrix

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8, 3.9]
 
     steps:
     - name: Checkout GitHub repository
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -1,8 +1,10 @@
 name: Build, test
 
 on:
-  schedule:
-    - cron: "0 2 * * 1,3,5"  # 02:00 UTC every Mon,Wed,Fri
+  push:
+    branches:
+      - main
+  pull_request:
 
 jobs:
   build:


jhlegarreta:
I think we are already in a position to make this happen/all aspects discussed in PR #46 are no longer a concern.

jhlegarreta:
Test failures should be fixed by the patch set in PR #174.

.github/workflows/build-test.yaml
```diff
@@ -1,16 +1,18 @@
 name: Build, test
 
 on:
-  schedule:
-    - cron: "0 2 * * 1,3,5"  # 02:00 UTC every Mon,Wed,Fri
+  push:
+    branches:
+      - main
+  pull_request:
```
josephmje, line 7: What do you think of limiting the GitHub Actions to only run when changes to the ipynb files are made? Since they are the target of the tests.

```suggestion
on:
  push:
    branches: main
    paths: '*.ipynb'
  pull_request:
    branches: main
    paths: '*.ipynb'
```

josephmje: Good point. What about limiting actions to ipynb files on push only and having no limits for PRs?

jhlegarreta: Not sure if I see the advantage/difference.  Can you please elaborate?

I had allegedly added push events so that when the `main` branch gets a merge commit it can get tested, so not sure if what you propose would allow to test the `main` branch when the requirements or Python versions change.

Note that I had first added `on: [push, pull_request]` and that triggered duplicate builds on this PR.

kaitj: > Note that I had first added on: [push, pull_request] and that triggered duplicate builds on this PR.

I suspect if you want to have both `push` and `pull_request` event triggers, you will also need to specify the type of PR event that triggers the workflow. I think pushing should also trigger a pull request action via `synchronize` (if branch is linked to an open PR) that is causing the duplicate build. 

As for limiting to ipynb files on push only to the `main` branch, I think I also am not seeing the full picture here.

josephmje: Disregard that. I think going forward, it may be better to use the `[skip ci]` commit message feature for commits that only affect the carpentries template. Does this sound like a better option rather than the earlier suggestion of triggering only when ipynb files change?

jhlegarreta: > > Note that I had first added on: [push, pull_request] and that triggered duplicate builds on this PR.
> 
> I suspect if you want to have both `push` and `pull_request` event triggers, you will also need to specify the type of PR event that triggers the workflow. I think pushing should also trigger a pull request action via `synchronize` (if branch is linked to an open PR) that is causing the duplicate build.

I'm not getting it. Please go ahead and amend the commit if you see a clear way for that.

> Disregard that. I think going forward, it may be better to use the [skip ci] commit message feature for commits that only affect the carpentries template. Does this sound like a better option rather than the earlier suggestion of triggering only when ipynb files change?

I am not convinced about it; the website check CI should run if changes to the episode markdowns are made provided that the check ensures that they comply with the style and the site builds.

josephmje: > I am not convinced about it; the website check CI should run if changes to the episode markdowns are made provided that the check ensures that they comply with the style and the site builds.

I think this is the only workflow that makes use of the `[skip ci]` flag. The other 2 workflows would still run despite using the flag.

jhlegarreta: > I think this is the only workflow that makes use of the [skip ci] flag. The other 2 workflows would still run despite using the flag.

:+1: You're right.

josephmje (approved):

josephmje:
I'm fine to merge this in.

jhlegarreta:
Merging.

jhlegarreta:
The merge commit has correctly triggered the actions on `main`.

PR #173, opened 2021-07-31
ENH: Install additional dependencies for `GitHub Actions` tasks

jhlegarreta:
Install additional dependencies for `GitHub Actions` tasks.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -40,6 +40,42 @@ jobs:
         pip install -r requirements.txt
         pip install pytest nbval
 
+    # Install additional dependencies
+    - name: Install additional dependencies
+      run: |
+
+        # Install CMake to be able to build ANTs
+        cd /
+        wget https://github.com/Kitware/CMake/releases/tag/v3.19.2
+        tar -xvzf cmake-3.19.2.tar.gz
+        rm -rf cmake-3.19.2.tar.gz
+        cd cmake-3.19.2
+        ./bootstrap -- -DBUILD_TESTING=OFF
+        make -j 8
+        make install
+
+        # Install FSL
+        cd /
+        wget https://fsl.fmrib.ox.ac.uk/fsldownloads/fslinstaller.py || echo "Failed to download FSL installer";
+        if [[ -f fslinstaller.py ]]; then
+          sudo python fslinstaller.py -d /usr/local/fsl --quiet || echo "Failed to install FSL";
+        fi
+
+        # Remove unnecessary FSL data
+        rm -rf /usr/local/fsl/*/data/atlases
+
+        # Instal ANTs
+        cd /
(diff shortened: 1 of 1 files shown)


jhlegarreta:
I inadvertently closed PR #41 and deleted the branch, so re-opening this as a new PR. I think we should try to figure this out to check the preprocessing notebook.

.github/workflows/build-test.yaml
```diff
@@ -40,6 +40,42 @@ jobs:
         pip install -r requirements.txt
         pip install pytest nbval
 
+    # Install additional dependencies
+    - name: Install additional dependencies
+      run: |
+
+        # Install CMake to be able to build ANTs
```
josephmje, line 47: Can an already compiled version of ANTS be installed? I think fmriprep may be using this in the Dockerbuild.

Something like:
```
    - uses: actions/cache@v2
      with:
        path: /opt/ants
        key: ants-v1
        restore-keys: |
          ants-v1
    - name: Install ANTS
      run: |
        if [[ ! -d "${ANTSPATH}" ]]; then
          sudo mkdir -p $ANTSPATH
          curl -sSL "https://dl.dropbox.com/s/gwf51ykkk5bifyj/ants-Linux-centos6_x86_64-v2.3.4.tar.gz" | sudo tar -xzC $ANTSPATH --strip-components 1
        fi
```

.github/workflows/build-test.yaml
```diff
@@ -40,6 +40,42 @@ jobs:
...
+        cd /
+        wget https://github.com/Kitware/CMake/releases/tag/v3.19.2
+        tar -xvzf cmake-3.19.2.tar.gz
+        rm -rf cmake-3.19.2.tar.gz
+        cd cmake-3.19.2
+        ./bootstrap -- -DBUILD_TESTING=OFF
+        make -j 8
+        make install
+
+        # Install FSL
```
josephmje, line 57: Likewise, can the ANTS and FSL steps be split? And add caching to both?

.github/workflows/build-test.yaml
jhlegarreta: Looks reasonable. Remains to be investigated.

.github/workflows/build-test.yaml
jhlegarreta: Looks reasonable. Remains to be investigated. Feel free to amend the commit if you know the sequence that works.

.github/workflows/build-test.yaml
josephmje: I'm sure this ANTs bit works since I've seen it used in other workflows. But it Installs 2.3.3 I think despite the filename being 2.3.4. I think it was compiled for Neurodebian so I'll see if there's a newer version.

jhlegarreta:
Superseded by #175. Closing.

PR #182, opened 2021-08-13
ENH: Use Exercise2 extension to synchronize solution and non-solution notebooks

### LINKED ISSUE #181: Combine solution and non-solution notebooks
I was playing around with the [Exercise2 extension](https://jupyter-contrib-nbextensions.readthedocs.io/en/latest/nbextensions/exercise2/readme.html) for Jupyter. It seems really easy to set up and I think it would help us combine the solution and non-solution notebooks into a single file. 

Here's a demonstration below. The solution can be revealed by clicking on a button. You can also include a cell below the solution whether the learner is meant to enter their own code. Here, you could still include fill in the blank style content and have pytest skip those cells when running the tests.

Peek 2021-08-12 16-52

WDYT? @kaitj @jhlegarreta

josephmje:
Closes #181 

This PR:
- combines the non-solution and solution notebooks into a single file
- updates the relative file paths
- uses the Exercise2 extension to reveal/hide the exercise solutions
- moves the `#NBVAL_SKIP` comment to a cell tag so it's hidden from the user


kaitj:
At a glance, content seems to be good - thanks for updating the notebooks. The DTI notebook is currently failing the tests due to a bad BIDS path.

jhlegarreta:
A few notes:
- Cross-referencing: https://github.com/carpentries-incubator/SDC-BIDS-dMRI/issues/181#issuecomment-898088473. Although having a single notebook would be nice, solutions need still to get tested. Is there a way to allow this in the tests?
- The changes to the notebooks are introducing changes that are inaccurate (e.g. check the documented parts, the relative path to the data in the trees; changes to the ipython version due to a local setting; some static figure paths do no longer refer to the good places, etc.)

josephmje:
> 
> 
> A few notes:
> 
> Although having a single notebook would be nice, solutions need still to get tested. Is there a way to allow this in the tests?

Yes, the solutions are still getting tested. The main update that this brings is that the solutions cells have an additional metadata tag that allows us to hide/reveal them. The cells that aren't getting tested are still the ones where we had the `#NBVAL_SKIP` comment from before.

> The changes to the notebooks are introducing changes that are inaccurate

I missed some of the path changes earlier but I think I've caught them all now.

jhlegarreta:
> Yes, the solutions are still getting tested. The main update that this brings is that the solutions cells have an additional metadata tag that allows us to hide/reveal them. The cells that aren't getting tested are still the ones where we had the #NBVAL_SKIP comment from before.

Then this is a :rocket:. I would not be convinced about bringing it otherwise.

Having a single notebook also brings the question of adding (and possibly hiding) the outputs (e.g. prints, images, etc.) generated by a cell: having the output of a cell is interesting to be able to know what that should give I think (I'm not sure whether it gets overwritten when executing the cell).

> I missed some of the path changes earlier but I think I've caught them all now.
There are still changes that are introduces without a good reason:
  - metadata/empty cells:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/182/files#diff-fdb730f6eed3e0bd9a7f71c381e8b9e097cc292b19dad6106cb63731c77599dbR31
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/182/files#diff-fdb730f6eed3e0bd9a7f71c381e8b9e097cc292b19dad6106cb63731c77599dbR80
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/182/files#diff-fdb730f6eed3e0bd9a7f71c381e8b9e097cc292b19dad6106cb63731c77599dbR128
and so forth.
  - Line breaks:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/182/files#diff-8bd28ca3bfa7adf6497ffb51a1fe41d86f989771d2baf3a2e699c4657e6b7fe9R428
  -  Version changes:
 https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/182/files#diff-fdb730f6eed3e0bd9a7f71c381e8b9e097cc292b19dad6106cb63731c77599dbR638

I ignore whether the version change is creating the metadata/empty cell changes; but I'd avoid adding those, since we will constantly be changing those whenever one has a slightly different version. Else, we should use the same version.

josephmje:
I think these issues that are being caught now also point to some differences between the solution and non-solution notebooks that we had already. The Review Notebook app gives a good diff so I'll check that it more detail.

> Having a single notebook also brings the question of adding (and possibly hiding) the outputs (e.g. prints, images, etc.) generated by a cell: having the output of a cell is interesting to be able to know what that should give I think (I'm not sure whether it gets overwritten when executing the cell).

For these notebooks, I opted to clear all cell output before pushing. This means that the outputs won't be rendered in the GitHub. I think this is necessary for this single notebook version because it would allow the learner to run the code without any prior expectation of what it should do. And if they would like to check that everything is correct, there's also the website render which contains all of the figure outputs.

> The README and index will need to be updated if we keep the solution and the paths to the notebooks change.

For the README, should we instead re-direct to the website links rather than the notebooks?

jhlegarreta:
> For these notebooks, I opted to clear all cell output before pushing. This means that the outputs won't be rendered in the GitHub. I think this is necessary for this single notebook version because it would allow the learner to run the code without any prior expectation of what it should do. And if they would like to check that everything is correct, there's also the website render which contains all of the figure outputs.

Correct :+1:. My mistake.

> For the README, it looks like we actually re-direct to the website links rather than the notebooks so it wouldn't be affected by this change. The markdown files are still the same.

True :+1. So no changes are required at that level.

jhlegarreta (approved):
:+1: .

kaitj (approved):

jhlegarreta:
I came across this:
https://jupyterbook.org/interactive/hiding.html

Have not investigated much, but maybe we now Jupyter supports hiding cells natively and we can drop the additional packages that were required for this?

PR #190, opened 2021-08-14
ENH: Increase preprocessing notebook cell execution timeout

jhlegarreta:
Increase preprocessing notebook cell execution timeout.

Fixes:
```
code/preprocessing::ipynb::Cell 5 XPASS (Previous cell timed out, ex...) [ 54%]
code/preprocessing::ipynb::Cell 6 XFAIL (Previous cell timed out, ex...) [ 63%]
code/preprocessing::ipynb::Cell 7 XPASS (Previous cell timed out, ex...) [ 72%]
code/preprocessing::ipynb::Cell 8 XPASS (Previous cell timed out, ex...) [ 81%]
code/preprocessing::ipynb::Cell 9 XPASS (Previous cell timed out, ex...) [ 90%]
code/preprocessing::ipynb::Cell 10 XPASS (Previous cell timed out, e...) [100%]
```

Raised for example in:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/3326949695?check_suite_focus=true#step:13:50

Reference documentation: https://jupyterbook.org/content/execute.html

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7]
     env:
       ANTSPATH: /opt/ants
       FSLPATH: /opt/fsl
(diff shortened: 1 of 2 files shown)


jhlegarreta:
I put 10 minutes for the relevant cells, but I do not know if that is enough or too long. Although this will depend on the VM features, maybe @kaitj @josephmje you can guesstimate better based on your experience with the notebook.

jhlegarreta:
Looks like the 10 minutes are either not enough or there is something else:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/190/checks?check_run_id=3329572110#step:13:50

jhlegarreta:
> We could bump up the time, but not sure if there is a time-limit for a job to run

Locally I tried to increase the timeout for the relevant cells in the preprocessing notebook (increased cells 4 and 5 -`topup` and `applytopup`-  to 50 minutes) without success:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/190/checks?check_run_id=3330545263#step:13:29

The thing is that the time spent on the whole notebook has not changed even if I increased first to 10 minutes, then 15, then 20, etc. So not sure the metadata that increases the allowed time property is being taken into account (even if 50 minutes falls short to the 55+ that you report).

I think there is something else going on that we are unable to identify.

> We may hit the limitations of github actions runners (both time and memory).

Time should not be an issue (we can go up to 6 hours). If memory is an issue, IMO we should use a smaller dataset.

PR #198, opened 2021-11-06
ENH: Increase text style consistency

jhlegarreta:
- ENH: Remove unused import statement.
- ENH: Avoid shadowing built-in type
- STYLE: Remove trailing whitespaces and add end of file newline
- DOC: Fix typos and improve wording
- STYLE: Use callouts for notes
- STYLE: Mark up a code statement
- STYLE: Make package names consistent
- STYLE: Use LaTeX syntax for exponential unit notation values


jhlegarreta:
A few notes:
- Using the links for all package mentions thanks to the `{% include links.md %}` statement would reduce maintenance in the Markdown files, but I guess that would not work in the Jupyter Notebooks, so I preferred to use backticks everywhere except where an explicit link was being required.
- As for the units, I guess we cannot do better (e.g. using LaTeX packages handling SI units).

_episodes/preprocessing.md
```diff
@@ -277,23 +277,23 @@ mv ../../data/ds000221/derivatives/uncorrected/sub-010006/ses-01/anat/sub-010006
 ~~~
 {: .language-bash}
 
-![T1w brain mask]({{ relative_root_path }}/fig/preprocessing/T1w_brainmask.png)
+![T1w brainmask]({{ relative_root_path }}/fig/preprocessing/T1w_brainmask.png)
```
josephmje, line 280: Should this be 2 words?

josephmje (approved):

_episodes/preprocessing.md
jhlegarreta: I think so, but there were single word instances dating from before I landed in the materials, and given that I'm not a native speaker, I chose to stick to the single word form. Do you want/prefer me to change them?

josephmje: I guess I've seen it both ways. Maybe keep it as one word as it corresponds to the filename better.

jhlegarreta: OK. Resolving and merging then. It can be changed at any moment in the future.

PR #200, opened 2021-11-08
ENH: Switch to `JupyterLab`

jhlegarreta:
Switch to `JupyterLab`: change mentions in documentation from `Jupyter
Notebooks` to `JupyterLab`.

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -2,13 +2,15 @@
 title: Setup
 ---
 
-This lesson uses [Jupyter Notebook] as a web-based interactive computational
-environment for learning. The `Jupyter Notebooks` can be run through [Binder],
-which builds a computing environment with all of the necessary software
-pre-installed. However, users may choose to run the notebooks locally. In that
-case, there are several pieces of software that must be installed. Although the
-main components required are `Python`-based, there are a few additional
-non-`Python` tools required.
+This lesson uses [JupyterLab] as a web-based interactive computational
+environment for learning. `JupyterLab` integrates a text editor, a terminal and
+`Jupyter Notebooks` through a single user-friendly interface, which will be
+used to run any code for the lesson. The `JupyterLab` instance and all of the
+necessary software have already been pre-installed on [Binder]. Users may
+choose to use this platform to get up and running much quicker, or may choose
+to run the notebooks locally. If running locally, there are several pieces of
+software that must be installed. Although the main components required are
+`Python`-based, there are a few additional non-`Python` tools required.
 
 Instructors should allocate some time to have all components installed.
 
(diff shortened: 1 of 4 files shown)


jhlegarreta:
Well, contrary to what I thought in https://github.com/carpentries-incubator/SDC-BIDS-dMRI/pull/193#issuecomment-962454077, it looks like `JupyterLab` is likely to have less or no impact on the notebook execution, since it is not in our requirements, and thus, not required to run the notebooks, and I now realize that the binder instance was already using it.

Anyways, we are more likely to get support I think.

setup.md
```diff
@@ -2,8 +2,8 @@
 title: Setup
 ---
 
-This lesson uses [Jupyter Notebook] as a web-based interactive computational
-environment for learning. The `Jupyter Notebooks` can be run through [Binder],
+This lesson uses [JupyterLab] as a web-based interactive computational
+environment for learning. The `Jupyter` notebooks can be run through [Binder],
 which builds a computing environment with all of the necessary software
 pre-installed. However, users may choose to run the notebooks locally. In that
 case, there are several pieces of software that must be installed. Although the
```
josephmje, line 9:
```suggestion
This lesson uses [JupyterLab] as a web-based interactive computational
environment for learning. `JupyterLab` integrates a text editor, terminal and `Jupyter Notebooks` through a single user-friendly interface, which will be used to run any code for the lesson. The `JupyterLab` instance and all of the necessary software have already been pre-installed on [Binder]. Users may choose to use this platform to get up and running much quicker, or may choose to run the notebooks locally. If running
locally, there are several pieces of software that must be installed. Although the
```

josephmje (commented):
Thanks for the transition @jhlegarreta! I just added some text to better introduce JupyterLab and Binder.

Also, I noticed during the lesson last Friday that the exercise/solutions package we're currently using only works with Jupyter Notebook. So we will also need to find an upgraded package to transition that as well.

jhlegarreta:
>  Also, I noticed during the lesson last Friday that the exercise/solutions package we're currently using only works with Jupyter Notebook. So we will also need to find an upgraded package to transition that as well.

:worried: I did not look into that. Looks like a rabbit hole then. Any idea if there is a fix/upgrade/alternative to the exercise/solution package?

setup.md
jhlegarreta: Done in 08512d6. Thanks @josephmje .

josephmje:
> 😟 I did not look into that. Looks like a rabbit hole then. Any idea if there is a fix/upgrade/alternative to the exercise/solution package?

I previously tried out the [Solutions extension](https://www.npmjs.com/package/@rmotr/jupyterlab-solutions) but I had trouble installing it.

Another workaround might be with [code folding](https://stackoverflow.com/questions/65532002/how-to-enable-code-folding-in-jupyter-lab)

josephmje (approved):

jhlegarreta:
> I previously tried out the Solutions extension but I had trouble installing it.
Another workaround might be with code folding

OK. You may want to open an issue to keep track of that.

In any case it looks like switching to `JupyterLab` is necessary to keep up with the `Jupyter` evolution.

I will merge once the CI will become green.

PR #204, opened 2022-02-22
ENH: Allow running visualization cells in CI workflow

jhlegarreta:
Allow running visualization cells in CI workflow by using the `xvfb`
display server tool.


josephmje (approved):
Thanks for this fix @jhlegarreta and sorry for the delay in reviewing it. I was wondering if you know whether Github actions allows you to actually see the results of the visualizations as well?

PR #205, opened 2022-03-05
ENH: Drop testing notebooks; only run them

jhlegarreta:
Drop testing notebooks; only run them. `pytest` and the `nbval` plugin
were being used to actually collect notebooks for testing, and to test
cells that output some value. However, `nbval`'s `timeout` flag to
explicitly set the cell execution time is not working properly, and most
`preprocessing` notebook cells were not being run.

Thus, this commit favors running all cells over testing the few cells
that print some output value.

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -101,7 +101,9 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-        pip install pytest nbval
+        pip install ipython
+        pip install jupyter
+        pip install nbconvert
 
     - name: Download data
       run: |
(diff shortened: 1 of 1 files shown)


jhlegarreta:
@kaitj  @josephmje The CIs are doing their job, and the preprocessing notebook seems to crash:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/5591600735?check_suite_focus=true#step:13:190

Any chance to have a look and fix that, please? Thanks.

A separate PR would be best, as I'd need to slightly clean the additions in here once we get all notebooks running.

kaitj:
Took a quick look but I suspect it has something to do with the way the command is called.

```\nmkdir -p ../data/ds000221/derivatives/uncorrected_topup_eddy/sub-010006/ses-01/dwi/work\n\n# Create an index file mapping the 67 volumes in 4D dwi volume to the pedir.txt file\nindx=""\nfor i in `seq 1 67`; do\n  indx="$indx 1"\ndone\necho $indx > ../data/ds000221/derivatives/uncorrected_topup_eddy/sub-010006/ses-01/dwi/work/index.txt\n\neddy_openmp \\\n  --imain=../data/ds000221/sub-010006/ses-01/dwi/sub-010006_ses-01_dwi.nii.gz \\\n  --mask=../data/ds000221/derivatives/uncorrected/sub-010006/ses-01/dwi/sub-010006_ses-01_brainmask.nii.gz \\\n  --acqp=../data/ds000221/derivatives/uncorrected_topup/sub-010006/ses-01/dwi/work/pedir.txt \\\n  --index=../data/ds000221/derivatives/uncorrected_topup_eddy/sub-010006/ses-01/dwi/work/index.txt \\\n  --bvecs=../data/ds000221/sub-010006/ses-01/dwi/sub-010006_ses-01_dwi.bvec \\\n  --bvals=../data/ds000221/sub-010006/ses-01/dwi/sub-010006_ses-01_dwi.bval \\\n  --topup=../data/ds000221/derivatives/uncorrected_topup/sub-010006/ses-01/dwi/work/topup \\\n  --out=../data/ds000221/derivatives/uncorrected_topup_eddy/sub-010006/ses-01/dwi/dwi \\\n  --repol\n'```

It looks like its adding the newline characters with how the notebook cells are run in the CI now. As far as I know, nothing has changed since we ran those tests to time these cells (and when we had tried it with `nbval`) so it should still run locally. I suspect moving everything to a single line would fix the failure.

jhlegarreta:
Thanks for the investigation Jason. Maybe defining the paths/filenames as variables will likely help to avoid having an endless, unreadable line, and hopefully make it digestible to the CI (?)

jhlegarreta:
I've just come across this:
https://pypi.org/project/nbmake/

Might be worthwhile trying: looks to accept/successfully apply the `timeout` value, avoids the need to convert the notebooks, and if we are lucky we might get back the cell-wise pass/fail reports. I'll have a look as time permits.

jhlegarreta:
@kaitj The same `eddy_openmp` step is failing in PRs #207 and #208 (I tried these two options to investigate which one provides the best report, allows to drop additional plugins, etc.)

PR #207, opened 2022-04-02
ENH: Prefer `nbmake` plugin for testing notebooks

jhlegarreta:
Prefer using the `nbmake` plugin over `nbval` for testing notebooks.

`nbval`'s `timeout` flag to explicitly set the cell execution time is
not working properly, and most `preprocessing` notebook cells were not
being run.

Documentation:
https://pypi.org/project/nbmake/

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -101,7 +101,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-        pip install pytest nbval
+        pip install pytest nbmake
 
     - name: Download data
       run: |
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Does not seem to provide a cell-wise report, but at least the `timeout` parameter is being correctly processed (although its 1h value seems to fall short for Python 3.9 and `topup`):
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/5800703673?check_suite_focus=true#step:13:58

alex-treebeard:
hi @jhlegarreta -- I maintain nbmake (sorry to intrude)

> Does not seem to provide a cell-wise report

Why would this help? Feature requests welcome 👍

jhlegarreta:
> hi @jhlegarreta -- I maintain nbmake (sorry to intrude)

@alex-treebeard No worries; we appreciate the time spent developing the plugin, and your interest on our use case.

> > Does not seem to provide a cell-wise report
> 
> Why would this help? Feature requests welcome +1

It's maybe a matter of preferences/nice-to-have feature, although, in practice, when a cell fails we still know which cell is the culprit.

A cell-wise report, as we had in the past using `pytest` and `nbval`, allows to have a clean report/summary, e.g.
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/5572292930?check_suite_focus=true#step:13:28

Also, without diving into the error trace, the report/summary shows the error type. In addition, `pytest` and `nbval` run all cells even if one fails, which `nbmake` does not seem to do at the moment (or at least I did not find the documentation to do so).

PR #208, opened 2022-04-02
ENH: Prefer `nclient` for running notebooks in testing

jhlegarreta:
Prefer using the `nbclient` plugin over `nbval` for running notebooks in
testing.

`nbval`'s `timeout` flag to explicitly set the cell execution time is
not working properly, and most `preprocessing` notebook cells were not
being run.

Documentation:
https://nbclient.readthedocs.io/en/latest/

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -101,7 +101,9 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install -r requirements.txt
-        pip install pytest nbval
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Much like the solution in #207, the `timeout` value seems to be working correctly.

But it is still falling short for `topup`:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/runs/5801333510?check_suite_focus=true#step:13:25

PR #215, opened 2022-10-22
ENH: Update actions versions to use `Node.js 16`

jhlegarreta:
Update actions versions to use `Node.js 16`.

Fixes:
```
Node.js 12 actions are deprecated.
For more information see:
https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
Please update the following actions to use Node.js 16: actions/cache, actions/checkout, actions/setup-python
```

raised for example in
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/actions/runs/3303231018

### DIFF
--- a/.github/workflows/build-test.yaml
+++ b/.github/workflows/build-test.yaml
@@ -20,7 +20,7 @@ jobs:
 
     steps:
     - name: Cache Ubuntu dependencies
-      uses: actions/cache@v2
+      uses: actions/cache@v3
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Warnings are gone:
https://github.com/carpentries-incubator/SDC-BIDS-dMRI/actions/runs/3303914423

PR #223, opened 2023-04-15
STYLE: Fix DTI episode radial diffusivity inline picture markup

jhlegarreta:
Fix DTI episode radial diffusivity inline picture markup.


jhlegarreta:
Should make the lambdas be inline:

fix_dmri_DTI_episode

jhlegarreta:
Build failures are unrelated to this patch set: the FSL installation step is timing out. I rerun the jobs after the first failure with the same result; the FSL installation does not take so long on a regular computer; we may be willing to find out why it takes so long here/find out how we can install it otherwise.

That should be addressed in a separate PR.

jhlegarreta:
The lambdas picture is now inline:

sdc_dti_lambdas_inline_fixed

PR #224, opened 2023-06-16
Update setup-python version

georgebisbas:
Updates the setup-python Github action to the latest major version


jhlegarreta (approved):
Thanks for doing this @georgebisbas.

jhlegarreta:
Again, as it has happened lately, the testing CI has timed out installing FSL. We definitely need to find a solution to this.

Merging anyways as the timeout is not related to this patch set, and the change is reasonable.

PR #238, opened 2023-08-13
STYLE: Allow rendering LaTeX math syntax: intro and preproc episodes

jhlegarreta:
Allow rendering LaTeX math syntax in introduction and preprocessing episodes.

Follow-up to commit 578080e5b971d397c9d7f99ff0469c318721d1cc.


jhlegarreta:
Not sure why the last GHA workflow actions had failed, as no apparent reason was displayed in the log, and why they succeed by re-triggering them.

Anyways, builds are passing, so merging.

PR #246, opened 2024-02-15
ENH: Use LaTeX equations in the DTI episode instead of inline images

jhlegarreta:
Use LaTeX equations in the DTI episode instead of inline images: the latter cause issues in the new Carpentries Workbench infrastructure.


jhlegarreta:
I am leaving the images just in case.

jhlegarreta:
Build failures are unrelated. Merging.

